PR #16, opened 2021-06-19
ocpp: handle unit conversion to kW, kWh

drc38:
@lbbrhzn, these are the last changes I was going to make while you're working on the structure.


lbbrhzn:
thanks!... i will start reworking the sructure now!

PR #20, opened 2021-06-22
major restructoring

lbbrhzn:
Major restructuring!
Some highlights:
- multi-platform support
- use ConfigFlow for configuration instead of configuration.yaml
- use devcontainers
- pre-commit checks

Todo: 
- add switch to enable/disable charging
- beautify the configuration flow
- add password option for ChargePoint registration
- add OptionFlow for selecting measurands
- ....


lbbrhzn:
@drc38, can you review this pull request?

drc38:
> @drc38, can you review this pull request?

Great! I'll have some time over the weekend to review.

drc38:
@lbbrhzn, with the new structure how are the sensors created associated with an instance of the `ChargePoint` class so that we could have more than one Charger instantiated with its own set of metrics/switches etc?

lbbrhzn:
> @lbbrhzn, with the new structure how are the sensors created associated with an instance of the `ChargePoint` class so that we could have more than one Charger instantiated with its own set of metrics/switches etc?

That's not implemented yet. This requires some additional work, since the sensors should be created after a new charge point connects.

drc38:
Conceptually will the `ChargePoint` need to become a HA entity and then in its setup/init add the sensors/switches etc?

lbbrhzn:
Yes, that's the idea. I think it should use the same forwarding strategy as in the __init__.py.

drc38:
In `sensors.py` you have used `MEASURANDS` from `const.py`
`for measurand in MEASURANDS:`
This should be using your `CONF_MONITORED_VARIABLES`

drc38:
Is there any reason to let the user change the `host`? If it is not `0.0.0.0`, `localhost` etc then the integration will fail is my expectation.

lbbrhzn:
A server can have multiple network interfaces. Using host=0.0.0.0 will bind to all IPv4 interfaces, which is sometimes not preferred from a security point of view, for instance when the server has both lan and wan interfaces. So we should allow the user to select a specific interface to bind to. This can be done by specifying the IP address of that interface. From what I under stand using '' (empty) would also bind to IPv6.

lbbrhzn:
I think we should use a more user friendly way to select the monitored variables, for instance by using checkboxes.

lbbrhzn:
Also, we should try to use an option flow as much as possible rather than a config flow for settings. That will allow the user to change settings without manually removing and re-adding the integration.

drc38:
Do these pre-commit warnings need to be resolved before I approve?
`UserWarning: `sections` setting includes INBETWEENS`
`UserWarning: W0503: Deprecated config options were used`

drc38 (approved):

PR #36, opened 2021-06-28
ocpp: add switches for basic functions

drc38:
@lbbrhzn, suggest a review of the coding as there is probably a better approach. Also with `start_transaction` there is a format violation with how the payload is formed. I haven't been able to work out why yet though.


drc38:
@lbbrhzn, if you think the approach to defining switches is suitable for the integration let me know and I can migrate it into your updated file structure.

lbbrhzn:
I'll take a look...

drc38:
Will send an updated PR based on new code

PR #61, opened 2021-07-08
Add websockets, ocpp module dependencies to test env

drc38:


drc38:
Not sure how to resolve this error `homeassistant.loader.IntegrationNotFound: Integration 'ocpp' not found.`

drc38:
@lbbrhzn , I think I've gone as far as I can. Still a few more issues to resolve with the testing scripts

lbbrhzn:
thnx drc38! Can we remove the test that is still failing?

.github/workflows/constraints.txt
```diff
@@ -1,5 +1,8 @@
-pip==21.1.3
+pip<=20.3
```
lbbrhzn, line 1: Is this <=20.3 coming from home assistant?

drc38: Yes it was

drc38:
@lbbrhzn , tests are now running without error but failing as there is not sufficient code coverage

lbbrhzn:
Ok, good enough for now!

PR #69, opened 2021-07-10
ocpp: migrate to use of enums more widely

drc38:


drc38:
@lbbrhzn, fully migrating to enums took a bit more work than expected. To help with testing but without affecting users  are you able to implement a tag for a stable release and also have bleeding edge version based on the latest commit for testing?

lbbrhzn:
Sure, I can try to make a tag (of the current main?) but I'm on holiday now for a few weeks, so my access to GitHub is a bit limited

PR #72, opened 2021-07-12
tests: create charge point test

drc38:


drc38:
@lbbrhzn , after merging #69 this should provide a basic test of the server-side charger code ie start, send meter data, and stop transaction. It should be fairly easy to extend out to cover others.

lbbrhzn:
Great!

Issue #77, opened 2021-07-16
Services are failing


lbbrhzn:
## Version of the custom_component
0.3.0

## Describe the bug

When toggling a switch, home assistant shows a pop-up:

```
Failed to call service switch/turn_on. Local variable 'resp' used before assignment.
```

lbbrhzn:
Assigning this to drc38

lbbrhzn:
Solved in v0.3.1

PR #75, opened 2021-07-16
ocpp: significant repo update

drc38:
@lbbrhzn, Combines the previous pull requests, adds custom services and testing of the bulk of api.py.  Should close #51 and #56

PR #81, opened 2021-07-16
ocpp: fix switch error #77

drc38:
Looks like this file got missed off #75


lbbrhzn (approved):

lbbrhzn:
Why do we need that specific sqlalchemy version?

PR #82, opened 2021-07-16
Revert "build(deps): bump sqlalchemy from 1.4.17 to 1.4.21 in /.github/workflows"

drc38:
Reverts lbbrhzn/ocpp#74

### DIFF
--- a/.github/workflows/constraints.txt
+++ b/.github/workflows/constraints.txt
@@ -4,5 +4,5 @@ black==21.7b0
 flake8==3.9.2
 reorder-python-imports==2.5.0
 websockets==9.1
-sqlalchemy==1.4.21
+sqlalchemy==1.4.17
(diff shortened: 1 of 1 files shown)


lbbrhzn:
Why do we require that specific sqlachemy version?

drc38:
`pytest-homeassistant-custom-component 0.4.2 depends on sqlalchemy==1.4.17`

lbbrhzn:
Is it possible to remove the requirement from requirements-dev.txt, and let pip auto detect this?

If not, we should add a comment to our requirements-dev.txt:

`# pytest-homeassistant-custom-component 0.4.2 depends on
sqlalchemy==1.4.17`

drc38:
Give it a go and see, I'm not familiar with how pip determines requirements

PR #85, opened 2021-07-17
ocpp: add websocket subprotocol support check

drc38:


lbbrhzn (approved):
Which subprotocols do we support?

drc38:
Only ocpp1.6

lbbrhzn:
Just wondering why this is needed. Doesn't websockets already check for the correct subprotocol by itself?

drc38:
I don't think so based on the comment in the example code here: https://github.com/mobilityhouse/ocpp/blob/master/examples/v16/central_system.py

PR #88, opened 2021-07-18
Push updates

drc38:
@lbbrhzn, progress towards updating metrics when data is received rather than awaiting the poll. There may be a better way to do this.

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -206,7 +206,9 @@ class CentralSystem:
         for ent in entity_registry.async_entries_for_device(er, dev.id):
             if SENSOR in ent.entity_id:
                 # _LOGGER.info("Entity id: %s updating", ent.entity_id)
-                entity_component.async_update_entity(self.hass, ent.entity_id)
+                self.hass.async_create_task(
+                    entity_component.async_update_entity(self.hass, ent.entity_id)
+                )
 
     def device_info(self):
         """Return device information."""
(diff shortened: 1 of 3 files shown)


lbbrhzn (commented):
I had implemented a different scheme (but this is now so far behind that I would need to start over): the sensors subscribe to metrics, and are called back when the metric value is updated.  Subscription should be done in the  `async_added_to_hass` callback, and unsubscribe on remove. See https://developers.home-assistant.io/docs/integration_fetching_data/ 

I've tried to copy the approach used in [zwave-js](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zwave_js/entity.py)


The sensor should return false on the 'should_poll' property.

drc38:
Because we already have callbacks when new data arrives eg `@on(Action...)` and multiple entities per device we know when sensors need updating without adding a listener on the sensor which is perhaps better but more complicated. I took the approach of updating all metrics as we don't have many entities rather than trying to update specific metrics based on data received.

Maybe rather than use the entity components update, it could be better to use `Entity.async_schedule_update_ha_state(force_refresh=True)` not sure though...

I kept polling on for now as a backstop in case something is missed in the push

lbbrhzn:
By making metric a class (with value and unit properties) that allows subscriptions, you can inform subscribers when there are changes.

Something like this (Pseudo code):

```python
ChargePoint:
self.metrics[current].value=new_value

Metric:
@value.setter
def value(self, new_value)
  self.value=new_value
  self.notify_subscribers()

Sensor:

add_to_hass():
  metric.subscribe(self.async_notify)

def async_notify(self)
  await self.async_write_ha_state()
```

drc38:
@lbbrhzn, your approach is more elegant. Up to you whether to merge this as an interim while you work on the subscriber approach.

lbbrhzn:
Ok, let's use this as an interim approach.

Issue #19, opened 2021-06-21
Support for 3 phases


Rep-Hueman:
Are phases also available?

"Power.Active.Import" is now (L3 ?!)

### Back office (OCPP) configuration key list for Alfen chargers

MeterValuesSampledData Sampled measurand(s) to be included in every meter value. The interval can be changed by 
changing the key "MeterValueSampleInterval". A combination, up to 9, of measurands is 
supported. The measurands can be combined by separating the value by a comma ","
Value example: 'Energy.Active.Import.Register,Voltage.L1-N,Current.Import.L1'

Supported values:
Energy.Active.Import.Register, Power.Active.Import, Current.Import, Voltage, 
Temperature, Current.Offered, Frequency, Power.Factor
Measurand(s) must be combined with the phase (separated by a dot ‘.’)

Supported phases:
Voltage meter values support phase configuration:
L1-N, L2-N, L3-N, L1-L2, L2-L3, L3-L1
Value example: 'Voltage.L1-N'

Current, power and power factor meter values support phase configuration:
L1, L2, L3
Value example: 'Current.Import.L1'

When the 'SmartChargingMode' (OCPP 1.5+) is beeing used, the following measurands are 
available:

Supported values:
Current.L1 ; Current.L2 ; Current.L3

lbbrhzn:
That's currently not available, because it is not part of the OCCP1.6 standard. Should not be that difficult to add though.

Rep-Hueman:
Log Home Assistant..

2021-06-21 18:14:32 INFO (MainThread) [ocpp] CH1Beheer: send [3,"220",{}]
2021-06-21 18:15:02 INFO (MainThread) [ocpp] CH1Beheer: receive message [2,"221","Heartbeat",{}]
2021-06-21 18:15:02 INFO (MainThread) [ocpp] CH1Beheer: send [3,"221",{"currentTime":"2021-06-21T16:15:02.648607"}]
[...]

PR #90, opened 2021-07-19
draft 3 phase processing of metrics

drc38:
#19 Hopefully this works as I can't test it other than through the test harness, but it should process L1 or L1-N data where present to extra state attributes. Phase to phase voltages are not processed L1-L2 etc as they can be calculated if needed.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,12 @@
+# These are supported funding model platforms
+
+github: [lbbrhzn, drc38]
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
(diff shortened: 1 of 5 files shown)


lbbrhzn:
Great! I assume this does not impact chargers that do not provide the phase related data, so it is safe to merge?

lbbrhzn (approved):

drc38:
> Great! I assume this does not impact chargers that do not provide the phase related data, so it is safe to merge?

Should be, it does make some changes to main metric logic though.  I am about to head off on holiday so if it does suggest reverting.

lbbrhzn:
Enjoy your holiday!

PR #93, opened 2021-07-20
Update current for 3 phase charger

drc38:
Switched to presenting overall current,  rather than average phase current for consistency between single and 3 phase chargers, so that overall charger real power = power.factor x current.import x voltage.

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -716,10 +716,7 @@ class ChargePoint(cp):
                     + value[Phase.l2.value]
                     + value[Phase.l3.value]
                 )
-                if sum > 0:
-                    self._metrics[metric] = round(sum / 3, 1)
-                else:
-                    self._metrics[metric] = round(sum, 1)
+            self._metrics[metric] = round(sum, 1)
             self._extra_attr[metric] = value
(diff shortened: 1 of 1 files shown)


custom_components/ocpp/api.py
```diff
@@ -716,10 +716,7 @@ def process_phases(self, data):
                     + value[Phase.l2.value]
                     + value[Phase.l3.value]
                 )
-                if sum > 0:
-                    self._metrics[metric] = round(sum / 3, 1)
-                else:
-                    self._metrics[metric] = round(sum, 1)
+            self._metrics[metric] = round(sum, 1)
```
lbbrhzn, line 719: Why round here? I assume it is for presentation purposes only, so maybe better do that in the sensor entity? 
We could extend the upcoming Metric class to have a format, besides value and unit.

lbbrhzn (approved):

custom_components/ocpp/api.py
drc38: Yes, only reason is presentation. Agree it would be better to handle format in the sensor. If sensor formats then other `round` functions can be removed from api also.

Issue #95, opened 2021-07-21
UnboundLocalError: local variable 'sum' referenced before assignment


Rep-Hueman:
## Version of the custom_component
 OCPP version  0.3.2  21-07-2021

## Configuration

```yaml

```

## Describe the bug

UnboundLocalError: local variable 'sum' referenced before assignment.

No problems with the test version for 3 phase - version 0.3.1 i test today.
## Debug log

```text

This error originated from a custom integration.

Logger: ocpp
Source: custom_components/ocpp/api.py:719
Integration: Open Charge Point Protocol (OCPP) (documentation, issues)
First occurred: 20:03:15 (1 occurrences)
Last logged: 20:03:15

[...]

PR #98, opened 2021-07-21
Update api.py

lbbrhzn:
Bugfix processing phase measurands. See #95

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -703,10 +703,7 @@ class ChargePoint(cp):
                     + value[Phase.l2_n.value]
                     + value[Phase.l3_n.value]
                 )
-                if sum > 0:
-                    self._metrics[metric] = round(sum / 3, 1)
-                else:
-                    self._metrics[metric] = round(sum, 1)
+                self._metrics[metric] = round(sum / 3, 1)
             if metric in [
                 Measurand.current_import.value,
                 Measurand.current_export.value,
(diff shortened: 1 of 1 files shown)


lbbrhzn:
@drc38, why are positive voltages treated differently from negative?

```Python

           if metric in Measurand.voltage.value:
                sum = (
                    value[Phase.l1_n.value]
                    + value[Phase.l2_n.value]
                    + value[Phase.l3_n.value]
                )
                if sum > 0:
                    self._metrics[metric] = round(sum / 3, 1)
                else:
                    self._metrics[metric] = round(sum, 1)
```

drc38:
> @drc38, why are positive voltages treated differently from negative?
> 
> ```python
>            if metric in Measurand.voltage.value:
>                 sum = (
>                     value[Phase.l1_n.value]
>                     + value[Phase.l2_n.value]
>                     + value[Phase.l3_n.value]
>                 )
>                 if sum > 0:
>                     self._metrics[metric] = round(sum / 3, 1)
>                 else:
>                     self._metrics[metric] = round(sum, 1)
> ```

There should not be negative voltages, and the >0 was to protect against a div by 0 error.

drc38 (approved):

lbbrhzn:
> ```python
>   if sum > 0:
>                     self._metrics[metric] = round(sum / 3, 1)
>                 else:
>                     self._metrics[metric] = round(sum, 1)
> ```

You’re never dividing by sum though. And if sum equals zero, the metric value will be zero.

Ok if I just change it to:

```python
self._metrics[metric] = round(sum / 3, 1)
```

drc38:
😄 good spotting, I was working too late. Please update

drc38:
Will you also update prior loop to reference `self._extra_attr`?

lbbrhzn:
> Will you also update prior loop to reference `self._extra_attr`?

I will keep that the same for now. I do think there is some room for simplification. We just want to average the voltages, and sum the currents, right?

Issue #101, opened 2021-07-23
Improve processing of measurands with phase components


lbbrhzn:
Some improvements are possible for processing measurands with phase components. Currently we only process currents and line-to-neutral voltages. We can extend this to support power and line-to-line voltages.

In general we should aim to provide the one-phase equivalent voltage, current, power and power factor whenever multiple phases are reported. 

When the voltages Vx-n are balanced (hope this is safe to assume?), we can do the following:

- Power and current values are summed.
- Power factors are averaged.
- Line-to-line voltages are divided by sqrt(3) to get line-to-neutral voltages.
- Reported and calculated line-to-neutral voltages are simply averaged.

See [here](https://powergenerators-ksa.com/en/three-phase-current-calculation/) for some background info.

_Originally posted by @lbbrhzn in https://github.com/lbbrhzn/ocpp/pull/98#issuecomment-885722169_

PR #102, opened 2021-07-23
101 improve phase processing

lbbrhzn:
See #101


lbbrhzn:
Can you take a look if these changes work? I have not yet been able to fully test them...

drc38:
Looking at the Tests error log it is saying there's an attribute error in your code when calling:
`Measurand.active_power_import.value`

drc38:
> Looking at the Tests error log it is saying there's an attribute error in your code when calling:
> `Measurand.active_power_import.value`

Should be: `power_active_import`

lbbrhzn:
Ok to merge?

PR #113, opened 2021-07-30
introduce the Metric class

lbbrhzn:


drc38:
@lbbrhzn , sorry I appear to have made a mistake trying to merge your metrics_class into my repo for testing. Are you able to revert this as I can't seem to?

lbbrhzn:
yes. I did:
'''
git reset dae06b
git  push --force origin
'''

custom_components/ocpp/api.py
```diff
@@ -793,23 +796,27 @@ def on_meter_values(self, connector_id: int, meter_value: Dict, **kwargs):
             processed_keys = []
             for idx, sv in enumerate(bucket[om.sampled_value.name]):
                 if m in sv and om.phase.value not in sv:
-                    self._metrics[sv[m]] = round(float(sv[om.value.value]), 1)
+                    self._metrics[sv[m]].value = round(float(sv[om.value.value]), 1)
                     if om.unit.value in sv:
                         if sv[om.unit.value] == DEFAULT_POWER_UNIT:
```
drc38, line 801: Not as a result of your changes but in reviewing I've noticed there should be a line here to set the units:
`self._metrics[sv[m]].unit = sv[om.unit.value]`

custom_components/ocpp/api.py
```diff
@@ -167,27 +168,30 @@ async def on_connect(self, websocket, path: str):
...
-            _LOGGER.info(f"Exception occurred:\n{e}")
+            _LOGGER.error(f"Exception occurred:\n{e}", exc_info=True)
+
         finally:
             self.charge_points[self.cpid].status = STATE_UNAVAILABLE
             _LOGGER.info(f"Charger {cp_id} disconnected from {self.host}:{self.port}.")
 
     def get_metric(self, cp_id: str, measurand: str):
         """Return last known value for given measurand."""
         if cp_id in self.charge_points:
```
drc38, line 179: Does this need `and (measurand in self.charge_points[cp_id]._metrics)`  as per lines 191-193?

custom_components/ocpp/api.py
```diff
@@ -167,27 +168,30 @@ async def on_connect(self, websocket, path: str):
...
-            return self.charge_points[cp_id].get_metric(measurand)
+            return self.charge_points[cp_id]._metrics[measurand].value
         return None
 
     def get_unit(self, cp_id: str, measurand: str):
         """Return unit of given measurand."""
         if cp_id in self.charge_points:
```
drc38, line 185: As per comment above

lbbrhzn: No. I've made metrics a `defaultdict`. As a result, any metric that is not (yet) defined will have default values (None)

custom_components/ocpp/api.py
lbbrhzn: No. I've made _metrics a `defaultdict`. As a result, any metric that is not (yet) defined will have default values (None)

custom_components/ocpp/api.py
lbbrhzn: ok, I'll add that

custom_components/ocpp/api.py
drc38: Useful to know as I was having lots of issues with missing dictionary values. 

Does that also mean `and (measurand in self.charge_points[cp_id]._metrics)` can be removed from lines 191-193?

drc38 (approved):
Looks good!

PR #137, opened 2021-09-03
New metric added to track websocket Reconnects

drc38:
This can be used to understand whether there may underlying network issues causing ocpp integration to become erratic/unstable


lbbrhzn:
looks good to me. merge?

PR #139, opened 2021-09-04
Change Reset switch to hard reset, trigger post_connect on boot and reset reconnects

drc38:
To help if HA/charger has a network error when starting

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -324,7 +324,6 @@ class ChargePoint(cp):
             self.status = STATE_OK
             await self.get_supported_features()
             if prof.REM in self._attr_supported_features:
-                await self.trigger_boot_notification()
                 await self.trigger_status_notification()
             await self.become_operative()
             await self.get_configuration(ckey.heartbeat_interval.value)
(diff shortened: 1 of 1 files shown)


custom_components/ocpp/api.py
```diff
@@ -566,8 +566,9 @@ async def stop_transaction(self):
             _LOGGER.debug("Failed with response: %s", resp.status)
             return False
 
-    async def reset(self, typ: str = ResetType.soft):
+    async def reset(self, typ: str = ResetType.hard):
         """Soft reset charger unless hard reset requested."""
```
lbbrhzn, line 570: Hard reset charger unless soft reset requested

custom_components/ocpp/api.py
```diff
@@ -867,8 +868,9 @@ def on_boot_notification(self, **kwargs):
             om.charge_point_serial_number.name, None
         )
 
-        asyncio.create_task(self.async_update_device_info(kwargs))
+        self.hass.async_create_task(self.async_update_device_info(kwargs))
         self.hass.async_create_task(self.central.update(self.central.cpid))
+        self.hass.async_create_task(self.post_connect())
```
lbbrhzn, line 873: are these tasks executed sequentially or in parallel?

drc38: In parallel, the Hass create tasks wraps the asyncio function. Figure its best to use Hass one.

drc38: Realised this could lead to loop, see new commit

lbbrhzn (approved):

PR #140, opened 2021-09-05
Set ping_timeout to None

drc38:
Avoids excessive reconnects

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -124,7 +124,11 @@ class CentralSystem:
         self = CentralSystem(hass, entry)
 
         server = await websockets.serve(
-            self.on_connect, self.host, self.port, subprotocols=self.subprotocol
+            self.on_connect,
+            self.host,
+            self.port,
+            subprotocols=self.subprotocol,
+            ping_timeout=None,
         )
         self._server = server
         return self


lbbrhzn:
Would it not be better to set the correct interval instead? Default seems to be 20 seconds, according to: https://websockets.readthedocs.io/en/stable/api/server.html, and it is set to 60 in [api.py](https://github.com/lbbrhzn/ocpp/blob/52c17b96e0ac0278758fea9880dbcc81508569e2/custom_components/ocpp/api.py#L331)

drc38:
The issue appears to be more fundamental than just the difference in timeouts, it loses connection  with 1006 [internal] which according to the docs https://websockets.readthedocs.io/en/stable/faq.html is only the result of a TCP error.  However the server should report 1011 on timeout according to this https://websockets.readthedocs.io/en/stable/api/server.html

lbbrhzn:
OCCP-J 1.6 specification discusses WebSocket Ping in relation to OCPP Heartbeat. 

> The WebSocket specification defines Ping and Pong frames that are used to check if the remote
> endpoint is still responsive. In practice this mechanism is also used to prevent the network operator
> from quietly closing the underlying network connection after a certain period of inactivity. This
> websocket feature can be used as a substitute for most of the OCPP Heartbeat messages, but cannot
> replace all of its functionality.
> An important aspect of the Heartbeat response is time synchronisation. The Ping and Pong frames
> cannot be used for this so at least one original Heartbeat message a day is recommended to ensure a
> correct clock setting on the Charge Point.

I dont see where we configure the heartbeat interval in our code. Do we rely on the charger to have this configured?

lbbrhzn:
> The issue appears to be more fundamental than just the difference in timeouts, it loses connection with 1006 [internal] which according to the docs https://websockets.readthedocs.io/en/stable/faq.html is only the result of a TCP error. However the server should report 1011 on timeout according to this https://websockets.readthedocs.io/en/stable/api/server.html

Strange. I guess for 1006 error setting the timeout to None has no impact?
Did you try turning on websocket debug logs to find out why the connection was closed?

drc38:
In theory it shouldn't,  as a 1006 should be due to a TCP/network  issue. The other approach would be to set charger ping interval to 0 but then the charger may not try to reconnect to server if ping/pong is lost, not sure...

lbbrhzn:
> Would it not be better to set the correct interval instead? Default seems to be 20 seconds, according to: https://websockets.readthedocs.io/en/stable/api/server.html, and it is set to 60 in [api.py](https://github.com/lbbrhzn/ocpp/blob/52c17b96e0ac0278758fea9880dbcc81508569e2/custom_components/ocpp/api.py#L331)

I'm a bit confused how this works. The client can send pings to the server, and the server can send pings to the client?
In that case each have there own set of intervals, and client side ping interval is not related to server side ping timeout.

drc38:
> > Would it not be better to set the correct interval instead? Default seems to be 20 seconds, according to: https://websockets.readthedocs.io/en/stable/api/server.html, and it is set to 60 in [api.py](https://github.com/lbbrhzn/ocpp/blob/52c17b96e0ac0278758fea9880dbcc81508569e2/custom_components/ocpp/api.py#L331)
> 
> I'm a bit confused how this works. The client can send pings to the server, and the server can send pings to the client?
> In that case each have there own set of intervals, and client side ping interval is not related to server side ping timeout.

I  am also... nothing suggests the server and client ping/pongs with different timing should not work. See https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers
The only fix I've found to 1006 disconnects so far is the above though. The only plausible guess I have is if server is at 20s and client is at 60s at some point they will coincide and according to implementation for multiple pings the receiver can choose to send only 1 pong. If the implementation expects 1 pong per ping this could result in a false disconnect... dunno as I can't seem to locate anything to suggest network goes down or logs from either server or client to provide detailed reason for the 1006

lbbrhzn:
> The only fix I've found to 1006 disconnects so far is the above though. The only plausible guess I have is if server is at 20s and client is at 60s at some point they will coincide and according to implementation for multiple pings the receiver can choose to send only 1 pong. If the implementation expects 1 pong per ping this could result in a false disconnect... dunno as I can't seem to locate anything to suggest network goes down or logs from either server or client to provide detailed reason for the 1006

That does not seem plausible to me. As far as I understand both server and client periodically send a ping, and both respond with a pong. When the pings coincide they are received on different hosts, so there is no collision or confusion, as long as the timeout for receiving the pong is less than or equal to the ping interval.

drc38:
Good point, as I say I'm stumped as to why setting it to None makes a difference

lbbrhzn:
You could try to log the communication with wireshark, that might shed some light on what is going on.

lbbrhzn:
Meanwhile, setting the timeout to None is fine with me... as long as it does not mean the connection will linger forever, and require some network reset or reboot on the server side.

lbbrhzn (approved):

drc38:
> Meanwhile, setting the timeout to None is fine with me... as long as it does not mean the connection will linger forever, and require some network reset or reboot on the server side.

All I could find from wireshark/logging is that the websocket close comes from the server side, after changing the timeout setting the charger diagnostics shows no disconnects for over 12h. If there was a real TCP issue its ping/pong should pick it up and log a disconnect. Server side has the 30s timeout on a message as a backstop.

PR #185, opened 2021-11-05
Add unit mapping

lbbrhzn:


drc38 (commented):
Suggest updating to new HA native measurement approach as part of this PR https://developers.home-assistant.io/docs/core/entity/sensor/

lbbrhzn:
Ok, I'll try that. Still need to nap OCPP units to HA units though.

drc38:
> Ok, I'll try that. Still need to nap OCPP units to HA units though.

From what I can see the others will not need explicit HA mapping. It is required for temp as HA converts to degC hence the error.

lbbrhzn:
> From what I can see the others will not need explicit HA mapping. It is required for temp as HA converts to degC hence the error.
It doesn't hurt to make the mapping explicit. Also, it's not only temperature. The reactive power and energy units (var, varh) have no direct equivalent in HA units, so must be mapped on W, Wh.

drc38:
> > It doesn't hurt to make the mapping explicit. 
I agree for metrics those that HA has a 1:1 mapping of units for.
> > The reactive power and energy units (var, varh) have no direct equivalent in HA units, so must be mapped on W, Wh.
For these I am suggesting it is better to map to `None`, as electrically VAh is a different unit to Wh. At some point HA will include and then they can be mapped also.

Hope that's clearer :)

PR #190, opened 2021-11-11
Update testing

drc38:


drc38:
@lbbrhzn not sure why the most recent pytest-homeassistant-custom-component version is now failing having resolved in 0.4.8... any ideas?

lbbrhzn:
Looks like the socket is blocked:

```
>       raise SocketBlockedError()
E       pytest_socket.SocketBlockedError: A test tried to use socket.socket.

/opt/hostedtoolcache/Python/3.9.8/x64/lib/python3.9/site-packages/pytest_socket.py:97: SocketBlockedError
```

Is the test using real sockets? If so, could the port be in use on the machine that is executing the tests?

lbbrhzn:
maybe add:
```
[pytest]
addopts = --allow-unix-socket --allow-hosts=127.0.0.1
```
to setup.cfg?

lbbrhzn:
Hmm, apparently not, still failing. I was able to upgrade pytest-homeassistant-custom-component  too 0.4.9 though, see  #189

Issue #236, opened 2021-12-13
v0.3.23 Problem with current


lbbrhzn:
### Discussed in https://github.com/lbbrhzn/ocpp/discussions/233

<div type='discussions-op-text'>

<sup>Originally posted by **McFranky** December 13, 2021</sup>
Since v0.3.23 the current_import sensor is 0A on my ABB Terra AC Wallbox. Before I got the actual current...</div>

McFranky:
In v0.3.24 the problem with no current always exists.
![grafik](https://user-images.githubusercontent.com/59682922/145927333-e3f2c742-0b7c-4bbd-81e7-ba34d3911dcb.png)

And now, I get no voltage! New Bug
![grafik](https://user-images.githubusercontent.com/59682922/145927437-6bd49afc-1c47-4966-a484-68e8657b7e1b.png)

McFranky:
Test Version 0.3.25 on ABB Terra AC
Now the current is to small!
![grafik](https://user-images.githubusercontent.com/59682922/146124428-51dde844-2c6d-419f-8167-729311625492.png)
And same by the voltage...
![grafik](https://user-images.githubusercontent.com/59682922/146124533-316578cb-1b03-4182-b4f1-4ea12c4b826b.png)

McFranky:
Test v0.3.26 on ABB Terra AC
![grafik](https://user-images.githubusercontent.com/59682922/146512844-47749391-a4d4-4d66-b12e-da71318583d1.png)
OK
![grafik](https://user-images.githubusercontent.com/59682922/146512977-1fdd3022-2c90-4352-b0b6-8055a2b8b8ec.png)
OK
![grafik](https://user-images.githubusercontent.com/59682922/146513065-12d13ef8-7f1f-44b5-9e5a-95215c0c4bbc.png)
OK

Super!! Great Job!

PR #238, opened 2021-12-14
Add test for 1p current

drc38:
May help with testing meter value processing, assertion test should pass when issue in #236 is fixed


lbbrhzn:
strange, should be passing now...

drc38:
> strange, should be passing now...

I think it works in order of PR, so is not running against the latest commit. Suggest committing to ensure it runs on latest code base.

lbbrhzn:
I merged it into the branch I had already made for the fix.

PR #239, opened 2021-12-14
Possible fix for #236

drc38:


lbbrhzn (approved):
was just debugging this, thanks for solving!

PR #241, opened 2021-12-14
Fix voltage when L1 reported, and L2,L3 are 0

drc38:
I think a change in the voltage section is required also based on "MeterValues" log in #233


lbbrhzn:
I agree with the fix for unit, but not with the change for line voltages.

If L2_N=0 you use L1_N? What if only L3_N is nonzero because you happen to be only charging with that phase?

An easy, consistent approach could be to average all defined and non-zero line voltages. This will allow any combination of one, two or three line voltages. 

Is it safe to assume all three phases are used when only l1_l2 is reported or should we also use the average of nonzero values approach there?

PR #242, opened 2021-12-15
Improve process phases

lbbrhzn:
another attempt to fix the mess...


lbbrhzn:
@drc38: I wonder if the unit conversion is still needed, since we do unit conversion in the sensor entity now.

drc38 (approved):
Averaging non-zeroes is a better approach 👍, I can't test at the moment as my system is out of action but I can't spot any issues with the code. On unit conversion I haven't had a chance to look at how the native unit and conversion in the entity works yet, but agree we should be able to remove from charger api.py

PR #247, opened 2021-12-18
Fix button typing

drc38:


drc38:
@lbbrhzn  looks like main issue was the tests.yaml python version. Is there any reason not to bump?

lbbrhzn:
I guess we should use the same version as home assistant.

drc38:
> I guess we should use the same version as home assistant.

My HA info says Python 3.9.7 but the sensor entity uses the same typing notation [here](https://github.com/home-assistant/core/blob/2021.12.2/homeassistant/components/sensor/__init__.py) whereas the typing documentation says | was introduced at 3.10... I'm confused 😕

drc38:
So looks like we're missing this import:
`from __future__ import annotations`
As discussed [here](https://www.python.org/dev/peps/pep-0563/)

lbbrhzn (approved):

Issue #220, opened 2021-12-06
2021.12.0 Beta deprecation warning


dreimer1986:
## Version of the custom_component

0.3.21

## Configuration

```yaml
            {
                "entry_id": "fe8d66744ea7798573a91ffdfb389031",
                "version": 1,
                "domain": "ocpp",
                "title": "central",
                "data": {
                    "host": "0.0.0.0",
                    "port": 9000,
                    "csid": "central",
                    "cpid": "evbox",
                    "meter_interval": 60,
                    "monitored_variables": ""
                },
                "options": {},
                "pref_disable_new_entities": false,
                "pref_disable_polling": false,
                "source": "user",
                "unique_id": null,
                "disabled_by": null
            },
```

## Describe the bug

2021.12.0 Beta reveals some new information on what Home Assistant want's to see coming from the addin and what will be deprecated soon. As the log entry asks to do, I inform you about the new information. I only get one warning right now from sensor.evbox_timestamp_config_response, but very likely sensor.evbox_timestamp_data_response will result in the same warning. No hurry though. Seems like the deprecation will not come before the February release.

There is another deprecation warning I can not yet tell where it comes from, but maybe it's interesting, too. 

## Debug log

```text

Logger: homeassistant.components.sensor
Source: components/sensor/__init__.py:376
Integration: Sensor (documentation, issues)
First occurred: 15:50:25 (1 occurrences)
Last logged: 15:50:25

sensor.evbox_timestamp_config_response is providing a string for its state, while the device class is 'timestamp', this is not valid and will be unsupported from Home Assistant 2022.2. Please report it to the custom component author.
```
The entry directly from log (There is a "Rejected while setting MeterValuesSampledData to" inside, which I get sometimes, too. Maybe interesting for you aswell):
```text
021-12-06 15:50:25 INFO (MainThread) [ocpp] EVB-P20482857: send [2,"4a191649-05cc-4c60-b677-1312b7db4518","GetConfiguration",{"key":["SupportedFeatureProfiles"]}]
2021-12-06 15:50:25 DEBUG (MainThread) [websockets.protocol] server > Frame(fin=True, opcode=<Opcode.TEXT: 1>, data=b'[2,"4a191649-05cc-4c60-b677-1312b7db4518","GetConfiguration",{"key":["SupportedFeatureProfiles"]}]', rsv1=False, rsv2=False, rsv3=False)
[...]

PR #248, opened 2021-12-18
Refactor Sensors for 2021.12+

drc38:
includes: 
- setting scale (how many dp) for metrics and handling dp formatting in sensor
- fix for #220

### DIFF
--- a/custom_components/ocpp/sensor.py
+++ b/custom_components/ocpp/sensor.py
@@ -33,7 +33,6 @@ async def async_setup_entry(hass, entry, async_add_devices):
     for metric in list(
         set(
             entry.data[CONF_MONITORED_VARIABLES].split(",")
-            + list(HAChargerDetails)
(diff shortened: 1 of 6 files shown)


custom_components/ocpp/sensor.py
```diff
@@ -31,26 +31,30 @@ async def async_setup_entry(hass, entry, async_add_devices):
...
-    for metric in list(HAChargerDetails):
+    for metric in list(HAChargerStatuses):
         SENSORS.append(
             OcppSensorDescription(
                 key=metric.lower(),
                 name=metric,
                 entity_category=EntityCategory.DIAGNOSTIC,
             )
         )
+    for metric in list(HAChargerDetails):
```
drc38, line 50: @lbbrhzn, these could also be DIAGNOSTIC instead of SYSTEM?

lbbrhzn (approved):

PR #254, opened 2021-12-20
Refactor Numbers for 2021.12+

drc38:


lbbrhzn:
Is this number version persistent over restarts of home assistant?

drc38:
> Is this number version persistent over restarts of home assistant?

Worth testing it, I'm not sure

drc38:
Added in RestoreEntity to make it persistent over restarts

lbbrhzn:
do you also resend the max current value when the charger reconnects?

drc38:
> do you also resend the max current value when the charger reconnects?

No it doesn't, it shouldn't be necessary and could cause problems if the websocket connection is not stable.

lbbrhzn (approved):

PR #264, opened 2021-12-26
Fix naming in numbers

drc38:


lbbrhzn (approved):

lbbrhzn:
Any idea why HACS load check is failing?

drc38:
> Any idea why HACS load check is failing?

No I don't,  maybe something missing in hacs.json? Also I haven't figured out why mixin RestoreEntity makes the number slider unavailable. Probably best to remove it for now.

PR #265, opened 2021-12-26
revert bump of pytest-homeassistant-custom-component

lbbrhzn:


lbbrhzn:
does not fix the HACS load check failure, so cancelling

PR #267, opened 2021-12-27
Update hacs.json

lbbrhzn:


drc38 (changes requested):
Add Button in domain also

PR #290, opened 2022-01-13
fix list validation error in test logic 

drc38:
refer #289


drc38:
Looks like it is due to an upstream change https://github.com/mobilityhouse/ocpp/commit/e213ee0d0f89ef8f3f244911a097062c8c69a01e

lbbrhzn:
@drc38 - can we go back to ocpp 0.10.0, at least until this is fixed?

PR #292, opened 2022-01-13
Add sampling config of meter when idle

drc38:
This PR adds regular sampling of the meter when it is not charging (ie idle) to the integration configuration. @lbbrhzn please add the Dutch translation


lbbrhzn (approved):
looks good to me!

PR #294, opened 2022-01-13
Fall back to ocpp=0.11.0

lbbrhzn:


custom_components/ocpp/manifest.json
```diff
@@ -4,14 +4,16 @@
...
+  "after_dependencies": [
+    "persistent_notification"
+  ],
   "codeowners": [
     "@lbbrhzn"
   ],
   "requirements": [
-    "ocpp>=0.8",
+    "ocpp==0.11.0",
     "websockets>=10.0"
```
drc38, line 17: Is it worth setting this to ==10.1 so that upstream changes don't affect us or can be tested first?

requirements_test.txt
```diff
@@ -1,4 +1,4 @@
 -r requirements_dev.txt
 pytest-homeassistant-custom-component==0.5.13
-ocpp==0.8.3
-websockets==10.1
\ No newline at end of file
+ocpp==0.11.0
```
drc38, line 3: Keep these two as >= so we can spot if an upstream issue causes testing to fail? If not then the manifest and requirements file can be bumped to the new version?

PR #312, opened 2022-01-18
add TxDefaultProfile to set charging rate

drc38:
for chargers that are incompatible with the ChargePointMaxProfile see #221


lbbrhzn (approved):
looks good to me!

PR #317, opened 2022-01-20
Measure connection latency

lbbrhzn:


drc38 (changes requested):
Looks good, if you add the measurement state class then statistics over time can be generated to see how stable the latency is

lbbrhzn:
> Looks good, if you add the measurement state class then statistics over time can be generated to see how stable the latency is

i'll do, thanks!

lbbrhzn:
Hmm, what SensorDeviceClass should I use? There is only timestamp, no timedelta.

drc38:
> Hmm, what SensorDeviceClass should I use? There is only timestamp, no timedelta.

Does it need one? Set to None?

custom_components/ocpp/api.py
```diff
@@ -774,6 +777,31 @@ async def _get_specific_response(self, unique_id, timeout):
...
+    async def measure_connection_latency(self):
+        """Measure the connection latency."""
+        try:
+            while True:
+                t0 = time.perf_counter()
+                pong_waiter = await self._connection.ping()
+                await pong_waiter
+                t1 = time.perf_counter()
+                latency = round(1000 * (t1 - t0))
+                _LOGGER.info(
```
drc38, line 789: suggest changing this to debug or commenting out before merging

custom_components/ocpp/api.py
```diff
@@ -774,6 +777,31 @@ async def _get_specific_response(self, unique_id, timeout):
 
         return resp
 
+    async def measure_connection_latency(self):
+        """Measure the connection latency."""
+        try:
+            while True:
+                t0 = time.perf_counter()
+                pong_waiter = await self._connection.ping()
```
drc38, line 785: use asyncio.wait_for(self._connection.ping(), timeout=20) to ensure the loop does not get stuck if there is no response to the ping

PR #327, opened 2022-01-23
tweak latency function

drc38:
@lbbrhzn I think it is best for the latency function not to close the connection or manage unexpected connection exceptions (leaving the main charger routine to do this), also if it does timeout to still log the 20000ms latency time.

### DIFF
--- a/tests/test_charge_point.py
+++ b/tests/test_charge_point.py
@@ -199,6 +199,8 @@ async def test_cms_responses(hass, socket_enabled):
             await asyncio.wait_for(
                 asyncio.gather(
                     cp.start(),
+                    cs.trigger_boot_notification(),
+                    cs.trigger_status_notification(),
                     test_switches(hass, socket_enabled),
                     test_services(hass, socket_enabled),
                     test_buttons(hass, socket_enabled),
--- a/tests/test_charge_point.py
+++ b/tests/test_charge_point.py
@@ -199,8 +199,8 @@ async def test_cms_responses(hass, socket_enabled):
             await asyncio.wait_for(
                 asyncio.gather(
                     cp.start(),
-                    cs.trigger_boot_notification(),
-                    cs.trigger_status_notification(),
+                    cs.charge_points[cs.cpid].trigger_boot_notification(),
+                    cs.charge_points[cs.cpid].trigger_status_notification(),
                     test_switches(hass, socket_enabled),
                     test_services(hass, socket_enabled),
                     test_buttons(hass, socket_enabled),
--- a/tests/test_charge_point.py
+++ b/tests/test_charge_point.py
@@ -318,7 +318,7 @@ class ChargePoint(cpclass):
     @on(Action.SetChargingProfile)
     def on_set_charging_profile(self, **kwargs):
         """Handle set charging profile request."""
-        return call_result.SetChargingProfilePayload(ChargingProfileStatus.accepted)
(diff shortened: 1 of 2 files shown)


tests/test_charge_point.py
```diff
@@ -318,7 +318,7 @@ def on_remote_stop_transaction(self, **kwargs):
     @on(Action.SetChargingProfile)
     def on_set_charging_profile(self, **kwargs):
         """Handle set charging profile request."""
-        return call_result.SetChargingProfilePayload(ChargingProfileStatus.accepted)
+        return call_result.SetChargingProfilePayload(ChargingProfileStatus.rejected)
```
lbbrhzn, line 321: I don't think this has anything to do with the latency measurement?!

lbbrhzn:
Hi @drc38, my motivation for closing the connection was this: when there is a timeout, either the network or the charger response is seriously degraded. Likely the charger will experience similar issues and try to reconnect. We don't want old connections to remain lingering, so we need to close at some point. Due to the latency measurement the normal ping/pong timeout mechanism of websockets will no longer work, so we need to do that manually.
 
How does this scenario play out with your changes?  How do we close a previous connection when the charger reconnects? Should we keep the connection open indefinitely if it doesn't?

tests/test_charge_point.py
lbbrhzn: I guess this is only for increasing test coverage.

lbbrhzn (approved):

tests/test_charge_point.py
drc38: Correct,  it was just to get test coverage >90%

drc38:
> Hi @drc38, my motivation for closing the connection was this: when there is a timeout, either the network or the charger response is seriously degraded. Likely the charger will experience similar issues and try to reconnect. We don't want old connections to remain lingering, so we need to close at some point. Due to the latency measurement the normal ping/pong timeout mechanism of websockets will no longer work, so we need to do that manually.
> 
> How does this scenario play out with your changes? How do we close a previous connection when the charger reconnects? Should we keep the connection open indefinitely if it doesn't?

My understanding is the client should initiate any close, including if the transport layer is degraded.   I don't think it is possible to have a lingering connection if the client reconnects (ie it has to close first). Leaving it to the client may prevent unnecessary closing where transport layer degrades but not for long enough for charger to close socket.

lbbrhzn:
My understanding is that either end can initiate a closing handshake in case of timeouts. See https://websockets.readthedocs.io/en/stable/topics/timeouts.html
Also note that the server can close at any time: See https://datatracker.ietf.org/doc/html/rfc6455#section-7.3

drc38:
> My understanding is that either end can initiate a closing handshake in case of timeouts. See https://websockets.readthedocs.io/en/stable/topics/timeouts.html Also note that the server can close at any time: See https://datatracker.ietf.org/doc/html/rfc6455#section-7.3

I guess it is more a design philosophy because as you say either end can close the connection. My wifi connection has medium signal strength and typical latency around 250ms, but at times goes to 3-4s and has gone to 20s without causing the charger to initiate a disconnect. The server could have dropped the connection but functionally it would not have made a difference as the transport layer went back to normal (possibly by the next ping/pong cycle). 

If you'd prefer the server to drop the connection I'm ok with that, perhaps change the function to  manage_connection_latency to make it clear it is actively managing the connection rather than a passive measurement.

lbbrhzn:
Let's revisit when the need arises.

PR #328, opened 2022-01-23
Slider unavailable workaround

drc38:
@lbbrhzn not sure if this will help but as we push api changes turning off polling might help

### DIFF
--- a/custom_components/ocpp/number.py
+++ b/custom_components/ocpp/number.py
@@ -90,6 +90,7 @@ class OcppNumber(RestoreEntity, NumberEntity):
         self._attr_max_value = self.entity_description.max_value
         self._attr_min_value = self.entity_description.min_value
         self._attr_step = self.entity_description.step
+        self._attr_should_poll = False
 
     async def async_added_to_hass(self) -> None:
         """Handle entity which will be added."""
@@ -131,3 +132,5 @@ class OcppNumber(RestoreEntity, NumberEntity):
             if resp is True:
                 self._attr_value = num_value
                 self.async_write_ha_state()
+        else:
+            self._attr_available = False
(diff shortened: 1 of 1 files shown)


lbbrhzn:
If I am not mistaken, we currently don't use push! It would require the entities to subscribe to change notifications.

I'm in favor of adding support for push though. We could implement it by add a subscription mechanism to the metric class. The metrics could also work as controls: when an entity changes the value of a metric, a notification is sent to a subscriber in the ChargePoint instance which will trigger an OCPP request to the charger. This way we could make slider always available even (when the charger is offline). We should just send the latest profile in the post_connect.
  
Longer term we could decouple the api from user interface by using MQTT to send the notifications. This would allow us to run the OCPP server separately from home assistant, for instance as an add-on.

drc38:
> If I am not mistaken, we currently don't use push! It would require the entities to subscribe to change notifications.
> 
> I'm in favor of adding support for push though. We could implement it by add a subscription mechanism to the metric class. The metrics could also work as controls: when an entity changes the value of a metric, a notification is sent to a subscriber in the ChargePoint instance which will trigger an OCPP request to the charger. This way we could make slider always available even (when the charger is offline). We should just send the latest profile in the post_connect.
> 
> Longer term we could decouple the api from user interface by using MQTT to send the notifications. This would allow us to run the OCPP server separately from home assistant, for instance as an add-on.



> If I am not mistaken, we currently don't use push! It would require the entities to subscribe to change notifications.
> 
> I'm in favor of adding support for push though. We could implement it by add a subscription mechanism to the metric class. The metrics could also work as controls: when an entity changes the value of a metric, a notification is sent to a subscriber in the ChargePoint instance which will trigger an OCPP request to the charger. This way we could make slider always available even (when the charger is offline). We should just send the latest profile in the post_connect.
> 
> Longer term we could decouple the api from user interface by using MQTT to send the notifications. This would allow us to run the OCPP server separately from home assistant, for instance as an add-on.

Sorry my reference to push was just for the slider, as in when its value is changed we push it to the api/charger. Applying a push framework more generally to the metrics sounds interesting, particularly for controls.

lbbrhzn (approved):

PR #334, opened 2022-01-30
monitor connection, close on timeout

lbbrhzn:
this my attempt on solving the ping buildup.

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -783,7 +783,6 @@ class ChargePoint(cp):
         self._metrics[cstat.latency_ping.value].unit = "ms"
         self._metrics[cstat.latency_pong.value].unit = "ms"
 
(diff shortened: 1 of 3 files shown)


drc38 (approved):
My suspicion is the ping latency will be close to 0 as `pong_waiter` will return as soon as websockets processes the request. I take it that's the idea in case an HA overload is causing an issue?

lbbrhzn:
I think even under high load, ping should return quickly, as long as the IP connection is up.
From websockets documentation: 
_Canceling ping() is discouraged. If ping() doesn’t return immediately, it means the write buffer is full. If you don’t want to wait, you should close the connection._
I assume a write buffers is emptied when the IP transmission is acknowledged. So this is a good indicator of network health. 
pong latency is influenced by network health, charger load, server load.

PR #336, opened 2022-01-30
Close old connection before switching to new

lbbrhzn:
Upon reconnect, make sure the old connection is closed before we start using the new connection

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -367,6 +367,7 @@ class ChargePoint(cp):
             await asyncio.sleep(2)
             await self.get_supported_features()
             if prof.REM in self._attr_supported_features:
+                await self.trigger_boot_notification()
                 await self.trigger_status_notification()
             await self.become_operative()
             await self.get_configuration(ckey.heartbeat_interval.value)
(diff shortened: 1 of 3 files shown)


drc38 (approved):

lbbrhzn:
@drc38, It looks like we no longer trigger a boot notification, is this not needed?
Or am I overlooking something... it seems to have test coverage.

drc38:
> @drc38, It looks like we no longer trigger a boot notification, is this not needed? Or am I overlooking something... it seems to have test coverage.

Not all chargers support it, I can't see that it is needed but it could be added back in.

lbbrhzn:
The boot notification contains useful device information, such as vendor, serial, firmware version. Without trigger it will only be sent after a reboot.  

Here's a scenario where I would like to trigger a boot notification: I tend to disable ocpp for firmware upgrades of my charger, which are done through the original management platform, and re-enable ocpp after the update. In this case the ocpp integration does not see a boot notification, so the firmware version is not updated. I propose to trigger the boot notification in post_connect if the Smart Charging profile is supported. Note that this should not trigger an actual reboot of the charger, just a resend of the information!

drc38:
Use the Remote profile, as that indicates support for trigger notifications rather than smart charging.

lbbrhzn:
Ah, you're right. The standard lists it incorrectly in the table, but the erratum fixes this.

drc38:
@lbbrhzn on 0.4.11 I get the ping latency metric but no pong metric? Is it the same for you?

lbbrhzn:
yes, i made a copy paste error in enums.py. will be fixed!

drc38:
That explains why it is also still reporting a similar latency time...it had me confused 🙃

PR #341, opened 2022-01-31
improve reconnect and trigger boot notification

lbbrhzn:
after adding trigger_boot_notification to post_connect, we could get into an infinite loop because the boot notification handler would create a new post_connect. This makes sense in case of a spontaneous reboot, but not if the boot notification was triggered. So we check for this, and only only re-do post connect if the boot notification was not triggered.

### DIFF
--- a/.devcontainer/configuration.yaml
+++ b/.devcontainer/configuration.yaml
@@ -3,7 +3,7 @@ default_config:
 logger:
   default: info
   logs:
-    custom_components.occp: debug
+    custom_components.ocpp: debug
 
 # If you need to debug uncomment the line below (doc: https://www.home-assistant.io/integrations/debugpy/)
 # debugpy:
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -153,7 +153,7 @@ class CentralSystem:
         self._server = server
         return self
 
-    async def on_connect(self, websocket: websockets.connection, path: str):
+    async def on_connect(self, websocket, path: str):
         """Request handler executed for every new OCPP connection."""
 
         try:
@@ -300,6 +300,8 @@ class ChargePoint(cp):
         self._requires_reboot = False
         self.preparing = asyncio.Event()
         self._transactionId = 0
+        self.triggered_boot_notification = False
+        self.received_boot_notification = False
         self._metrics = defaultdict(lambda: Metric(None, None))
         self._metrics[cdet.identifier.value].value = id
         self._metrics[csess.session_time.value].unit = TIME_MINUTES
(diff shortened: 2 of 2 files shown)


drc38 (commented):
Should there be a `finally:`  before line 821?

drc38:
I think you also need to set `return_exceptions=True` on gather and then cancel the gather on exception to stop the monitor_connection task in the `start` and `reconnect` try blocks.

lbbrhzn:
yes, something like that is needed. But I've read somewhere that cancelling  gather does not automatically cancel the underlying tasks. So you need to do something like this:
https://python.tutorialink.com/how-to-cancel-all-remaining-tasks-in-gather-if-one-fails/

drc38:
My understanding is if return_exceptions is True then cancelling gather will cancel the other tasks, but I haven't confirmed with a piece of test code. The difference being you have to test the result for an Exception. https://hynek.me/articles/waiting-in-asyncio/

drc38:
> My understanding is if return_exceptions is True then cancelling gather will cancel the other tasks, but I haven't confirmed with a piece of test code. The difference being you have to test the result for an Exception. https://hynek.me/articles/waiting-in-asyncio/

Ignore this, you are right it will only work for finite tasks as the result from gather is only returned after all tasks have completed successfully or thrown an exception (which may happen but there is no guarantee).  I agree option 2 looks cleaner.

drc38:
Another option worth considering is removing the closing from the monitor_connection function, and simply propagate the error back to the calling try block (start and reconnect), after which the connection will close in the finally statement. The logging and closing can then be handled there eg

In monitor_connection:
```
except Exception as e:
    raise e
```

In start/reconnect:
```
except asyncio.TimeoutError:
            _LOGGER.debug(f"Timeout in connection '{self.id}'")
        except websockets.exceptions.ConnectionClosed as connection_closed_exception:
            _LOGGER.debug(
                f"Connection closed to '{self.id}': {connection_closed_exception}"
            )
        except Exception as other_exception:
            _LOGGER.error(
                f"Unexpected exception in connection to '{self.id}': {other_exception}",
                exc_info=True,
            )
```

lbbrhzn:
got something that seems to work now, please review.

drc38 (changes requested):
I'm getting the following error:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/websockets/legacy/server.py", line 293, in handler
    await self.ws_handler(self, path)
  File "/config/custom_components/ocpp/api.py", line 203, in on_connect
    f"Charger {cp_id} disconnected from {self.host}:{self.port} websocket={websocket.id}."
AttributeError: 'WebSocketServerProtocol' object has no attribute 'id'
```

lbbrhzn:
Strange, works for me. What is your websockets version?

lbbrhzn:
i'll release as beta now, so others can check if it works

drc38:
> Strange, works for me. What is your websockets version? Should be there, see https://websockets.readthedocs.io/en/stable/reference/server.html#websockets.server.WebSocketServerProtocol

That's odd somehow HA reverted to websockets 9.1 although I have been on 10.1 for sometime... I had to go into bash and upgrade to get 10.1 back again. This may be the reason:
```
volvooncall 0.9.1 requires websockets<=8.1, but you'll have websockets 10.1 which is incompatible.
vallox-websocket-api 2.8.1 requires websockets<10.0,>=9.1, but you'll have websockets 10.1 which is incompatible.
```

PR #355, opened 2022-02-02
Avoid spurious reconnects

drc38:
Adds a counter to avoid forcing a reconnect if the connection is back online by the next 2 tries


lbbrhzn (changes requested):
Please make the timeout_counter configurable. This will also aid in testing.
Can you add tests for this functionality?

drc38:
I've added the basic tests, I'm unsure how to go about replacing the websockets pong response in a fixture to cause it to raise the timeout exception though

lbbrhzn:
I thing you can disable the websocket connection by manually setting state to CLOSING.
This will prevent pongs being sent.

lbbrhzn (changes requested):
WEBSOCKET_CLOSE_TRIES is confusing. 
I would prefer WEBSOCKET_PING_TRIES.

lbbrhzn (approved):
great!

PR #362, opened 2022-02-03
add config option to skip schema validation

lbbrhzn:


drc38 (approved):
Lgtm

N3rdix:
Couldn’t find any info what this parameter does, will this be added to the docs?
At least it sounds interesting from my perspective 😁

lbbrhzn:
Hi @N3rdix, the ocpp package we use verifies that the communication adheres to the schema defined in the standard.
In case there is a deviation it throws an exception and the connection is dropped. This configuration option allows the user to disable the validation.

N3rdix:
Hey @lbbrhzn,
thanks for the explanation und great work by the way!

PR #365, opened 2022-02-03
trigger status notifications for all connectors

lbbrhzn:


custom_components/ocpp/api.py
```diff
@@ -490,16 +490,18 @@ async def trigger_boot_notification(self):
...
-        )
-        resp = await self.call(req)
-        if resp.status == TriggerMessageStatus.accepted:
-            return True
-        else:
-            _LOGGER.warning("Failed with response: %s", resp.status)
-            return False
+        """Trigger status notifications for all connectors."""
+        return_value = True
+        for id in range(0, 1):
```
drc38, line 495: Use `self._metrics[cdet.connectors.value].value` here instead?

drc38:
Perhaps in  the `@on(Action.StatusNotification)` use the extra attributes of the connector status metric to handle statuses and errors for more than 1 connector?

lbbrhzn:
Not sure what you mean. The functionality to store status and error code separately for  main (id=0) and connector (id>0) is already in.

drc38:
> Not sure what you mean. The functionality to store status and error code separately for main (id=0) and connector (id>0) is already in.

Is it for only 1 (ie 0 and 1) connector or will it handle say 0, 1 and 2?

lbbrhzn:
We could support more than one connector, but to display the info in separate sensors we would need to know how many there at setup, or create the additional sensors on the fly, which is messy.

Alternatively we could just keep the two sensors for connector=0 and connector=1 as is, and store additional info for other connectors in extra_attributes.

custom_components/ocpp/api.py
lbbrhzn: It does need to be known, so the number of connectors should be retrieved before sending the triggers.

drc38:
> So, let's just keep the two sensors for connector=0 and connector=1 as is, and store additional info for other connectors in extra_attributes.

Agree 👍

PR #375, opened 2022-02-06
improve MeterValues handing with transaction id

lbbrhzn:


custom_components/ocpp/api.py
```diff
@@ -1023,8 +1038,9 @@ def on_meter_values(self, connector_id: int, meter_value: dict, **kwargs):
                         self._metrics[measurand].value = float(value) / 1000
                         self._metrics[measurand].unit = HA_POWER_UNIT
                     elif unit == DEFAULT_ENERGY_UNIT:
-                        self._metrics[measurand].value = float(value) / 1000
-                        self._metrics[measurand].unit = HA_ENERGY_UNIT
+                        if not transaction_matches:
```
drc38, line 1041: Assume this is true when transactionId=0, but why is the if statement required?

custom_components/ocpp/api.py
```diff
@@ -1000,21 +1008,28 @@ def average_of_nonzero(values):
...
     def on_meter_values(self, connector_id: int, meter_value: dict, **kwargs):
         """Request handler for MeterValues Calls."""
-        self._metrics[csess.transaction_id.value].value = kwargs.get(
-            om.transaction_id.name, 0
-        )
+
+        transaction_id: int = kwargs.get(om.transaction_id.name, 0)
+
+        transaction_matches: bool = False
+        if transaction_id > 0:
```
drc38, line 1015: Can this if be removed because if transaction_id=0 if on 1016 will still evaluate to false?

PR #382, opened 2022-02-08
Fix bug in transactionId match

drc38:


drc38:
@lbbrhzn any ideas why in testing of the `send_start_transaction` the charge point's `active_transactionId` is not being updated to the Id provided by the cms server and instead reporting 1000 which is set at init?

lbbrhzn (approved):

lbbrhzn (approved):

lbbrhzn:
looks good to me, but tests are not passing yet !

drc38:
@lbbrhzn, should be good to merge now. Some extra tests can now be added eg function to send clock meter data

lbbrhzn:
Looking good, nice test coverage! 
Ok to merge now?

PR #392, opened 2022-02-12
Improve post_connect logic

drc38:
If post_connect always triggers a charger issue eg disconnect this may cause continuous reconnects

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -323,6 +323,7 @@ class ChargePoint(cp):
         self.active_transaction_id: int = 0
         self.triggered_boot_notification = False
         self.received_boot_notification = False
+        self.post_connect_success = False
         self.tasks = None
         self._metrics = defaultdict(lambda: Metric(None, None))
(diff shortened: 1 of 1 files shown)


tests/test_charge_point.py
```diff
@@ -268,6 +268,8 @@ async def test_services(hass, socket_enabled):
         # use same id to ensure metrics populated
         cp = ChargePoint("CP_1_test", ws)
         cp.accept = False
+        cs.charge_points[cs.cpid].received_boot_notification = False
+        cs.charge_points[cs.cpid].post_connect_success = False
```
drc38, line 272: @lbbrhzn, not sure why these statements do not fully test the if statements above, any ideas?

lbbrhzn: looks like the NotImplementedError is not thrown.

drc38: I'm not sure how the  NotImplementedError should work, there is also the exceptions.py code which is not called during testing but references throwing an error configuring the charger.

lbbrhzn (approved):

PR #399, opened 2022-02-15
Fix/streamline german translation

N3rdix:
I just saw the good work of @dreimer1986 and fixed some smaller typos on the fly. Additionally I streamlined some wordings to be better readable.


dreimer1986:
No objections. A few are indeed typos, some more a case of different taste, but all in all I like it. ^^

lbbrhzn:
reviewed and accepted. Thanks!

PR #419, opened 2022-03-01
add id_tag to session energy extra_attr

drc38:


custom_components/ocpp/api.py
```diff
@@ -1089,6 +1089,9 @@ def on_meter_values(self, connector_id: int, meter_value: dict, **kwargs):
                 self._metrics[csess.session_energy.value].value = float(
                     self._metrics[DEFAULT_MEASURAND].value or 0
                 ) - float(self._metrics[csess.meter_start.value].value)
+                self._metrics[csess.session_energy.value].extra_attr = {
+                    cstat.id_tag.name: self._metrics[cdet.identifier.value].value
+                }
```
lbbrhzn, line 1094: I don't think this is correct.
self._metrics[cdet.identifier.value].value is set to the charge point identity in line 329, and is never updated.
It will always be equal to the charge point identity.
You should use self._metrics[cstat.id_tag.value].value, which is set at the start of a transaction.

lbbrhzn (changes requested):
see comments

custom_components/ocpp/api.py
drc38: Good spot, I copied the wrong id.

PR #437, opened 2022-03-12
add authorization list

lbbrhzn:
Add support for authorization list in configuration.yaml.

Example:

```yaml
ocpp:
  default_authorization_status : 'Invalid'
  authorization_list: 
    - id_tag : '123456789'
      name:  'accepted tag'
      authorization_states: 'Accepted'
    - id_tag : '987654321'
      name: 'expired tag'
      authorization_status: 'Expired'
    - id_tag : '987654321'
      name: 'blocked tag'
      authorization_status: 'Blocked'
```

### DIFF
--- a/.devcontainer/configuration.yaml
+++ b/.devcontainer/configuration.yaml
@@ -1,19 +1,20 @@
 default_config:
 
 ocpp:
+  default_authorization_status: 'Invalid'
   authorization_list:
     - id_tag: '12345678'
       name: 'My tag'
-      status : Accepted
+      authorization_status : Accepted
     - id_tag: 'CAFEBABE'
       name: 'Some other tag'
-      status: Blocked
+      authorization_status: Blocked
     - id_tag: 'DEADBEEF'
       name: 'Old tag'
       status: Expired
     - id_tag: '12341234'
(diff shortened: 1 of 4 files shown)


lbbrhzn:
still needs some more test coverage.

drc38 (commented):
I haven't had a chance to test, but a few comments looking through the code:
 - is it worth adding the parentId to the configuration.yaml list?
 - is it worth defining anything other than accepted in the configuration.yaml as the default is either invalid (or could be blocked) and expired requires timing logic (eg tag authorised for 1h use)?

lbbrhzn:
>     * is it worth adding the parentId to the configuration.yaml list?
That might be useful to group tags, for instance different tags that belong to the same person. Should be easy to add.
  
>     * is it worth defining anything other than accepted in the configuration.yaml as the default is either invalid (or could be blocked) and expired requires timing logic (eg tag authorised for 1h use)?
I just made it flexible enough to support all allowed responses. A charger could display a message with the reason that tag is not accepted.  As I understand it, providing an expiration time is mainly used for a localAuthList. We probably want to set LocalAuthListEnabled=False and handle authorization in home assistant.

PR #443, opened 2022-03-16
skip subprotocol validation when configured

lbbrhzn:


custom_components/ocpp/api.py
```diff
@@ -177,7 +179,9 @@ async def on_connect(
         self, websocket: websockets.server.WebSocketServerProtocol, path: str
     ):
         """Request handler executed for every new OCPP connection."""
-
+        if self.config.get(CONF_SKIP_SCHEMA_VALIDATION, DEFAULT_SKIP_SCHEMA_VALIDATION):
```
drc38, line 182: @lbbrhzn, won't this logic mean the setup of the ChargePoint is also skipped rather than just skipping the subprotocol verification section?

PR #471, opened 2022-03-31
default to Core profile for malformed responses

drc38:


lbbrhzn:
looks good. Should we log a warning when an unknown FeatureProfile is received?
Should be possible using elifs.

lbbrhzn:
Add a test with an empty featureprofile?

drc38:
> Add a test with an empty featureprofile?

I've added a test in for that but I haven't figured out why it doesn't run yet.

drc38:
Should be good to merge now

PR #506, opened 2022-04-30
add secure option

drc38:
@lbbrhzn unfortunately I cannot test whether this will work as the manufacturer has locked users out from its configuration system. A certificate can be created by following these instructions https://community.home-assistant.io/t/certificate-authority-and-self-signed-certificate-for-ssl-tls/196970


lbbrhzn:
Looks good. Would it be possible to add a test with ssl enabled?

drc38:
> Looks good. Would it be possible to add a test with ssl enabled?

In theory it should be possible, not sure how you create the required certs on the test machine for server and client though. Any ideas?

lbbrhzn:
Use default self signed certs, disable certificate verification.

PR #563, opened 2022-07-10
Update new device

victorsanva:
new device working and tested


victorsanva:
working this product without problem

lbbrhzn:
thanks for the update!

PR #599, opened 2022-08-15
ProjectEV(Atess) chargers return empty responses to config messages b…

archerjm:
…ut are smart enabled chargers.  Support these by making the default type core plus smart in the case fo an empty message and protect the set_charging_rate calls against the empty responses that these chargers generate.


drc38:
Defaulting in the code to "Core with Smart Charging" will cause issues in the integration for those chargers that don't support the Smart Charging profile. I suggest adding this to the Configuration Flow as a tickbox option eg "Add manual support for Smart Charging". Defaulting to a stack level of 0 when a response is blank looks fine to me.

drc38:
@archerjm try the latest release and feel free to do another PR if needed

Issue #510, opened 2022-05-10
Support for Sync EV Charger


FingalP:
**Is your feature request related to a problem? Please describe.**
The ocpp integration doesn't work with my Sync EV charger. It claims to be OCPP 1.6 compliant but it isn't fully. I get a few different errors/issues:
- Sync EV charger doesn't recognise `SupportedFeatureProfiles` - it returns it as an unknown key, rather than `[""]` as the code expects for a charger which only supports Core.
- `ChangeAvailability` requests just time out. By default, this causes an error and the whole thing reconnects, which makes it unusable.
- Some metrics are set with the key as the connector_id, which is an integer. Homeassistant fails to log this because it expects keys to be strings.

**Describe the solution you'd like**
I've made a fork https://github.com/FingalP/ocpp which has workarounds. I'm not sure whether this is something which should be merged into the main repo because it's supporting a charger which isn't fully OCPP compliant, but it would be convenient for me if it was.

**Describe alternatives you've considered**
I have contacted Sync EV to see if they will update their code to make it fully OCPP compliant. No response so far.

**Additional context**
I work for [Carbon Coop](https://carbon.coop/), integrating with EV chargers for the [Powershaper Flex](https://flex.powershaper.io/) project and the [REScoop VPP](https://www.rescoopvpp.eu/) project.

drc38:
The SupportedFeatureProfiles and connector_id changes look fine to me. I think there is an easier way to deal with change availability, in `post_connect` move `await self.set_availability()` to the "nice to have" section. It should still timeout but then reconnect successfully. 

Also I'm interested in whether you have an update on the OpenDSR project, it'd be good to do something similar in NZ?

FingalP:
Hi @danharvey 

I went to the "Admin Panel" page and enter the PIN (which is just `admin`).

Then you can modify the websocket address (they call it Web Address) in the Backend Details tab. A restart is required after that.

Hope that helps!

PR #612, opened 2022-09-06
Add "force smart charging" option to configuration flow

birneee:
Add "force smart charging" option to configuration flow
Suggested by [drc38](https://github.com/drc38) in https://github.com/lbbrhzn/ocpp/pull/599#issuecomment-1225090294, https://github.com/lbbrhzn/ocpp/discussions/474#discussioncomment-2577256
Fixes https://github.com/lbbrhzn/ocpp/issues/510, https://github.com/lbbrhzn/ocpp/pull/599 (at least partially)
Fixes https://github.com/lbbrhzn/ocpp/discussions/474


drc38:
Please update the tests to include the new config option and fix linting, cheers

drc38 (commented):
Best to add your test in this section `# test ocpp rejection messages sent from charger to cms`
The dummy charger will return a blank response to the get FeatureProfiles configuration and it can be asserted the cms returns SmartCharging when the config flow option is set to True

sreichholf:
I do have a suggestion for this (I do own a ABB Terra AC so I am actually affected by this)

Wouldn't cases like these be a typical scenario for "vendor/model specific quirks" as uses e.g. in ZHA?
Most typical users barely understand options like this and start a "trial and error session" until everything does what they want it to.

At least for known cases it should be possible to detect device specific quirks and handle them within the code and in return improve the "plug 'n play" experience for users.

drc38:
> I do have a suggestion for issues like this (I do own a ABB Terra AC so I am actually affected by this).
> 
> Wouldn't cases like this one be a typical scenario for "vendor/model specific quirks" as used e.g. in ZHA? Most typical users barely understand options like this and start a "trial and error session" until everything does what they want it to. Plus there is a real "threat" that the list of "device specific options" gets longer and longer.
> 
> At least for known cases it should be possible to detect device specific quirks and handle them within the code and in return improve the "plug 'n play" experience for users.

In terms of the Ocpp HA integration it was intended to be completely generic to the standard and not to really accommodate device specific quirks. However over time it has become clear a number of manufacturers are not fully compliant with the standard or there are some "grey" areas eg websockets, supporting a feature but not reporting that it does. 

To assist others with device specific quirks or configuration this can be documented in the docs via a PR or the repo forked to create a device/manufacturer specific integration. Hopefully over time most quirks will be addressed through firmware upgrades.

sreichholf:
I don't mean to be to pessimitic, but having worked with communication "standards" in multiple industries (primarily EDI in the automotive sector Automotive and pretty much every single standard ever used in the "Enterntainment Sector" (DVB, HDMI, CeC) this seems like an unachievable goal.

I do understand and honor the concept/wish though and agree that user-side configuration is fine when documented properly.

PR #622, opened 2022-09-15
Master

archerjm:
Updates to allow for the empty messages that the ProjectEV(ATESS) charger returns.  Also added in additional diagnostic variables that this smart charger produces.


custom_components/ocpp/enums.py
```diff
@@ -34,6 +34,10 @@ class HAChargerStatuses(str, Enum):
     firmware_status = "Status.Firmware"
     reconnects = "Reconnects"
     id_tag = "Id.Tag"
+    current_import = "Current.Import"
```
drc38, line 37: These changes are not needed, please remove from the PR. See the upstream enums  repository https://github.com/mobilityhouse/ocpp

drc38:
Please fix the linting and add tests for the new code

custom_components/ocpp/const.py
```diff
@@ -39,7 +39,7 @@
 DEFAULT_MAX_CURRENT = 32
 DEFAULT_PORT = 9000
 DEFAULT_SKIP_SCHEMA_VALIDATION = False
-DEFAULT_FORCE_SMART_CHARGING = False
+DEFAULT_FORCE_SMART_CHARGING = True
```
drc38, line 42: The default should not be changed in the code base. Simply remove ocpp integration from HA, reinstall and select the force smart charging option.

archerjm: This change wasn't meant o be part of the pull request was just so users of the current ProjectEV version didn't have to change this setting after pulling your updates into the code.

archerjm:
Closing pull request until issues are resolved.

PR #752, opened 2023-01-24
Add path to SSL certificate

moorgrove:
Add ability to specify the path to where the certificate and key files are stored.


drc38 (changes requested):
Are you able to confirm the secure connection works? When I added it I had no ability to test it with my charger.
Please fix the linting and testing errors, and add details to the ReadtheDocs on how to use the secure setup.

moorgrove:
> Are you able to confirm the secure connection works? When I added it I had no ability to test it with my charger.
> Please fix the linting and testing errors, and add details to the ReadtheDocs on how to use the secure setup.

I have a CTEK Chargestorm 2 connected with wss:// which is a SSL connection as I understand it. It reports data and works with "secure connection" applied in the OCPP Add On settings.

I'm sorry for all the commits. I still learning git, linting, pre-commit and to run tests. I see that I still have "not successful" for codecov/patch. Is it mandatory to add non global variables in the test process. And do you have a recommendation what information I would need to add to .readthedocs.yaml?

PR #915, opened 2023-09-25
temp fix for jsonschema

drc38:


lbbrhzn (commented):
Pinning the jsonschema to a particular version may easily create conflicts with other integrations. Is there any other way to fix this?

lbbrhzn:
The tests are failing with the change. Any idea why?

lbbrhzn:
Approving even with failed tests to work around a problem in upstream package.

drc38:
> The tests are failing with the change. Any idea why?

No idea, suggest updating to python 3.11, HA and pytest to latest. Then try and solve any issues from there.

PR #919, opened 2023-09-26
Added detection between session and main meter energy values

nlindn:
Added detection between session and main meter energy values as discussed in #366.

According to OCPP spec, if there is a transaction id the values are for that session and if there is no transaction id the values are from main meter. For example ABB Terra AC sends separate energy used for the session and the charger lifetime energy usage.

Now if there is a transaction id with the energy value, it is saved to entity "xxx.Session" (example: "Energy.Active.Import.Register.Session")

If there is no transaction id with the energy value, it is saved to entity "xxx" (example: ""Energy.Active.Import.Register)

This way the energy values between the sessions and main meters are kept separated so that they don't mix in the homeassistant history and graphs. Also, if the charger gives the energy used in the session it is not calculated duplicated.

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -11,7 +11,7 @@ import time
 
 from homeassistant.components.persistent_notification import DOMAIN as PN_DOMAIN
 from homeassistant.config_entries import ConfigEntry
-from homeassistant.const import STATE_OK, STATE_UNAVAILABLE, TIME_MINUTES
+from homeassistant.const import STATE_OK, STATE_UNAVAILABLE, STATE_UNKNOWN, TIME_MINUTES
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers import device_registry, entity_component, entity_registry
 import homeassistant.helpers.config_validation as cv
@@ -245,6 +245,12 @@ class CentralSystem:
             return self.charge_points[cp_id]._metrics[measurand].value
         return None
 
+    def del_metric(self, cp_id: str, measurand: str):
+        """Set given measurand to None."""
+        if cp_id in self.charge_points:
+            self.charge_points[cp_id]._metrics[measurand].value = None
+        return None
+
     def get_unit(self, cp_id: str, measurand: str):
         """Return unit of given measurand."""
         if cp_id in self.charge_points:
@@ -353,6 +359,7 @@ class ChargePoint(cp):
         self.received_boot_notification = False
         self.post_connect_success = False
         self.tasks = None
+        self._charger_reports_session_energy = False
         self._metrics = defaultdict(lambda: Metric(None, None))
         self._metrics[cdet.identifier.value].value = id
         self._metrics[csess.session_time.value].unit = TIME_MINUTES
@@ -1073,6 +1080,30 @@ class ChargePoint(cp):
 
         transaction_id: int = kwargs.get(om.transaction_id.name, 0)
 
+        # If missing meter_start or active_transaction_id try to restore from HA states. If HA
+        # does not have values either, generate new ones.
+        if self._metrics[csess.meter_start.value].value is None:
+            value = self.get_ha_metric(csess.meter_start.value)
+            if value is None:
+                value = self._metrics[DEFAULT_MEASURAND].value
+            else:
+                value = float(value)
+                _LOGGER.debug(
+                    f"{csess.meter_start.value} was None, restored value={value} from HA."
+                )
+            self._metrics[csess.meter_start.value].value = value
+        if self._metrics[csess.transaction_id.value].value is None:
+            value = self.get_ha_metric(csess.transaction_id.value)
+            if value is None:
+                value = kwargs.get(om.transaction_id.name)
+            else:
+                value = int(value)
+                _LOGGER.debug(
+                    f"{csess.transaction_id.value} was None, restored value={value} from HA."
+                )
+            self._metrics[csess.transaction_id.value].value = value
+            self.active_transaction_id = value
+
         transaction_matches: bool = False
         # match is also false if no transaction is in progress ie active_transaction_id==transaction_id==0
         if transaction_id == self.active_transaction_id and transaction_id != 0:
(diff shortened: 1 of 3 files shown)


drc38:
Thanks, I take it ABB have not released a new firmware to conform to 3.9 of the errata then? 
https://github.com/mobilityhouse/ocpp/blob/master/docs/v16/ocpp-1.6-errata-sheet.pdf

nlindn:
> Thanks, I take it ABB have not released a new firmware to conform to 3.9 of the errata then? https://github.com/mobilityhouse/ocpp/blob/master/docs/v16/ocpp-1.6-errata-sheet.pdf

Yeah, the spec 3.9 is not implemented yet in Terra AC. The Sample.Periodic response has the session energy started from 0 for every session.

`receive message [2, "2503015", "MeterValues", 
	{"connectorId": 1, "transactionId": 1695585075, "meterValue": 
		[{"timestamp": "2023-09-24T21:40:03.000Z", "sampledValue": 
			[{.................................,
			{"value": "1868", "context": "Sample.Periodic", "format": "Raw", "measurand": "Energy.Active.Import.Register", "unit": "Wh"},
			..................................}]
		}]
	}
]`

And the Sample.Clock sends the main meter values.

`receive message [2, "2778188", "MeterValues",
	{"connectorId": 1, "meterValue": 
		[{"timestamp": "2023-09-24T21:45:00.000Z", "sampledValue": 
			[{"value": "927435", "context": "Sample.Clock", "format": "Raw", "measurand": "Energy.Active.Import.Register", "location": "Inlet"}]}]}]`

If other chargers on the market are sending the main meter value not starting from 0 as suggested in the 3.9 of the errata. Then the default assumption in this pull request implementation is maybe not the best. Do we want the default assumption to be that the energy values does not start from zero for transactions and if 3.9 of the errata is not yet implemented, then do some workarounds to stop the mixing of values that is happening now?

drc38:
>If other chargers on the market are sending the main meter value not starting from 0 as suggested in the 3.9 of the errata. Then the default assumption in this pull request implementation is maybe not the best. Do we want the default assumption to be that the energy values does not start from zero for transactions and if 3.9 of the errata is not yet implemented, then do some workarounds to stop the mixing of values that is happening now?

That would be better, it seems to only be the ABB units with the issue. Also any reason to change from `Energy.Session` as it will break any reporting users have setup etc?

nlindn:
> That would be better, it seems to only be the ABB units with the issue. Also any reason to change from `Energy.Session` as it will break any reporting users have setup etc?

Changed the logic of the fix as discussed and kept the entitie names unchanged to avoid breaking setups.

drc38:
@nlindn we shouldn't need to maintain a list of charger models in the integration that do not support errata 3.9. Can you rework to remove that requirement please.

drc38:
Why did you drop the test case in the latest commits?

nlindn:
> @nlindn we shouldn't need to maintain a list of charger models in the integration that do not support errata 3.9. Can you rework to remove that requirement please.

From the meterStart value on startTransaction it can be detected if the errata 3.9 is fulfilled or not. Problem is that if homeassistant is restarted during a charging session the meterStart value is not available after the reboot for the still ongoing session. Any ideas for a good solution for getting the meterStart value for ongoing session if startTransaction is missed?

drc38:
> > @nlindn we shouldn't need to maintain a list of charger models in the integration that do not support errata 3.9. Can you rework to remove that requirement please.
> 
> From the meterStart value on startTransaction it can be detected if the errata 3.9 is fulfilled or not. Problem is that if homeassistant is restarted during a charging session the meterStart value is not available after the reboot for the still ongoing session. Any ideas for a good solution for getting the meterStart value for ongoing session if startTransaction is missed?

It should be fairly rare for a HA restart to occur during a charge session. I wouldn't worry about it.

drc38:
Great work 🙏.  Please add `jsonschema==4.19.0` to requirements_dev.txt as I suspect that is a reason for the tests not running. Cheers

tests/test_charge_point.py
```diff
@@ -310,6 +314,54 @@ async def test_services(hass, socket_enabled):
...
+                asyncio.gather(
+                    cp.start(),
+                    cp.send_boot_notification(),
+                    cp.send_authorize(),
+                    cp.send_heartbeat(),
+                    cp.send_status_notification(),
+                    cp.send_security_event(),
+                    cp.send_firmware_status(),
+                    cp.send_data_transfer(),
+                    cp.send_start_transaction_meter_start_zero(),
```
drc38, line 337: Lines 330 to 336 are not required as they've already been tested

tests/test_charge_point.py
```diff
@@ -310,6 +314,54 @@ async def test_services(hass, socket_enabled):
...
+                    cp.send_meter_err_phases(),
+                    cp.send_meter_line_voltage(),
+                    cp.send_meter_periodic_data(),
```
drc38, line 340: Same with 338-339, this is slowing test and causing timeout

drc38 (changes requested):

nlindn:
Removed duplicated tests. Timeout of 30sec was still not enough so increased the timeout for test_cms_responses in test_charge_point.py to 60sec.

drc38:
> Removed duplicated tests. Timeout of 30sec was still not enough so increased the timeout for test_cms_responses in test_charge_point.py to 60sec.

Nice one, just linting to fix and take a look at codecov report to see if you can increase test coverage.

drc38 (approved):

drc38:
@nlindn, testing the changes on my unit and it breaks the logic for a normal charger
image

Please review

custom_components/ocpp/api.py
```diff
@@ -1099,8 +1130,25 @@ def on_meter_values(self, connector_id: int, meter_value: dict, **kwargs):
...
+                            # Charger reports Energy.Active.Import.Register directly as Session energy for transactions
+                            self._charger_reports_session_energy = True
+                        if (
+                            transaction_matches
+                            and self._charger_reports_session_energy
+                            and measurand == DEFAULT_MEASURAND
+                            and connector_id
+                        ):
+                            self._metrics[csess.session_energy.value].value = (
+                                float(value) / 1000
```
drc38, line 1144: @nlindn this should only be /1000 if unit == DEFAULT_ENERGY_UNIT

custom_components/ocpp/api.py
```diff
@@ -1099,8 +1130,25 @@ def on_meter_values(self, connector_id: int, meter_value: dict, **kwargs):
...
+                            )
+                            self._metrics[csess.session_energy.value].extra_attr[
+                                cstat.id_tag.name
+                            ] = self._metrics[cstat.id_tag.value].value
+                        elif (
+                            transaction_matches or self._charger_reports_session_energy
+                        ):
                             self._metrics[measurand].value = float(value) / 1000
                             self._metrics[measurand].unit = HA_ENERGY_UNIT
```
drc38, line 1153: Same applies here

tests/test_charge_point.py
```diff
@@ -310,6 +373,43 @@ async def test_services(hass, socket_enabled):
```
drc38, line 373: adding `assert float(cs.get_metric("test_cpid", "Energy.Active.Import.Register")) == float(
        1101.452
    )` here will show error with current code

nlindn:
> @nlindn, testing the changes on my unit and it breaks the logic for a normal charger

Yeah, the logic has some bugs that passed the tests but reports values wrong on errata 3.9 chargers.  Should I open a new pr for fixes or what is the best way to do it?

drc38:
Open a new PR

PR #954, opened 2023-10-28
adopt new integration blueprint

lbbrhzn:
adopt new integration blueprint, to allow more recent version of pytest-homeassistant-custom-component
see https://github.com/ludeeus/integration_blueprint

With a recent version of pytest-homeassistant-custom-component some test failures pop-up, which need to be solved.


lbbrhzn:
@drc38, do you know what to to to fix the following error?

`ERROR tests/test_charge_point.py::test_cms_responses - Failed: Lingering task after test <Task pending name='Task-695' coro=<ChargePoint.send_stop_transaction() running at /home/runner/work/ocpp/ocpp/tests/test_charge_point.py:1103> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[gather.<locals>._done_callback() at /opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/asyncio/tasks.py:764]>
`

drc38:
I assume it is this line:
1102: `while self.active_transactionId == 0:`
try commenting it out.

lbbrhzn:
that does not solve the problem, we then get an error with unknown transaction id=0

drc38:
> that does not solve the problem, we then get an error with unknown transaction id=0

How about replacing line below
1103: `await asyncio.sleep(1)`
With `return`

lbbrhzn:
Also does not work. Maybe you can help me out through codespace:

https://prod.liveshare.vsengsaas.visualstudio.com/join?138371B71D13139074524B061EC019DD6C9F

lbbrhzn:
Added @chrisK824 as a reviewer. He has experience in making an [OCPP charging point simulator](https://github.com/chrisK824/ocpp_charging_point_simulator).

drc38:
> Also does not work. Maybe you can help me out through codespace:
> 
> https://prod.liveshare.vsengsaas.visualstudio.com/join?138371B71D13139074524B061EC019DD6C9F

Tried to access link but suspect time difference is a problem, I'm in NZ. Should have some time at weekend to take a look at this.

drc38:
Looks like the timeout cannot be relied upon to close tasks, solution committed is not very elegant but gets around the issue.

drc38 (approved):

lbbrhzn (commented):
it may not be elegant, but it works!

PR #968, opened 2023-11-06
autodetect supported measurands

lbbrhzn:
add autodetection of supported measurands.

If this works well, we can remove configuration of the measurands from the config flow.


drc38 (approved):
I can't test with my charger for another week, but the changes lgtm

PR #975, opened 2023-11-16
new option flows & tags

lbbrhzn:
- added new option flows to allow reconfiguration of central system and chargepoint point parameters.
- added config and option flow for tags
- added support multiple charge points per central system. 

NB. work in progress (TODO: get the tests working, add sensors for energy usage per tag)

@drc38  I am considering to allow only a single central system instance. Assuming we fully support multiple charge points per central system, do you see any reason to allow more than one central system instance?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,11 +1,16 @@
 {
-  "python.linting.pylintEnabled": true,
-  "python.linting.enabled": true,
-  "python.pythonPath": "/usr/local/bin/python",
   "python.analysis.extraPaths": [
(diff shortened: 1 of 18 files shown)


drc38:
Hi @lbbrhzn I agree it makes sense to only have 1 central server. I installed the draft and appreciate it is work in progress so in case it is useful here a few things I found:

- after adding a charge point device when I click on configure it throws an error saying the device already exists
- two instances of the charge point device were added, not sure if it was a hangover from prior integration
- my charger would not successfully connect, I suspect due to the issue with loading the charger device, stopped around line 228 of `api.py`

Issue #1053, opened 2024-01-21
Support for more than one OCPP client


munterkalmsteiner:
**Is your feature request related to a problem? Please describe.**
Currently, only one charge point is supported by the central system. This leads to an issue in setups where a load balancer serves as a proxy for the charge point (see the Garo Entity Pro [issue](https://github.com/lbbrhzn/ocpp/issues/1043)).

**Describe the solution you'd like**
The setup of the component should allow for creating more than one charge point.

**Describe alternatives you've considered**
Currently, both central station and charge point are setup in one step.

Alternative 1: the config flow is a two-step process, i.e. first defining the central station, and then, in the second step, the charge point parameters. In the second step, there could a check box that, when checked, will repeat the second step to setup another charge point.

Alternative 2: the config flow could ask first what should be created (central station or charge point) and in the second step, depending on the choice of device, the parameters are configured. Only one central station device is allowed, hence when one is already created, the config flow defaults to creating a charge point device.

PR #1054, opened 2024-01-21
Support more than one charge point

munterkalmsteiner:
Addresses #1053.

This is a draft. Tests and updated documentation missing.

### DIFF
--- a/custom_components/ocpp/__init__.py
+++ b/custom_components/ocpp/__init__.py
@@ -14,13 +14,12 @@ from .api import CentralSystem
 from .const import (
     CONF_AUTH_LIST,
     CONF_AUTH_STATUS,
-    CONF_CPID,
+    CONF_CHARGE_POINTS,
     CONF_CSID,
     CONF_DEFAULT_AUTH_STATUS,
     CONF_ID_TAG,
     CONF_NAME,
     CONFIG,
-    DEFAULT_CPID,
     DEFAULT_CSID,
     DOMAIN,
     PLATFORMS,
@@ -79,14 +78,15 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         model="OCPP Central System",
     )
 
-    """ Create Charge Point Device """
-    dr.async_get_or_create(
-        config_entry_id=entry.entry_id,
-        identifiers={(DOMAIN, entry.data.get(CONF_CPID, DEFAULT_CPID))},
-        name=entry.data.get(CONF_CPID, DEFAULT_CPID),
-        model="Unknown",
-        via_device=((DOMAIN), central_sys.id),
-    )
+    """ Create Charge Point Devices """
+    for cp_id in entry.data.get(CONF_CHARGE_POINTS):
+        dr.async_get_or_create(
+            config_entry_id=entry.entry_id,
+            identifiers={(DOMAIN, cp_id)},
+            name=cp_id,
+            model="Unknown",
+            via_device=((DOMAIN), central_sys.id),
+        )
 
     hass.data[DOMAIN][entry.entry_id] = central_sys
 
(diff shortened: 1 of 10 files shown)


lbbrhzn:
hi @munterkalmsteiner, 
I started adding functionality for multiple charge points a few months ago already, but was never able to finish it due to lack of time. You can find my code in the new_option_flows branch. My approach was to allow the user to create three different types of entities: Central System, Charge Point, and Tag.  The properties could also be edited after creation, using OptionFlows. Maybe you an take a look and salvage what you find useful?

munterkalmsteiner:
@lbbrhzn I see! There is definitely something I can peek/reuse from, in particular since I'm new to the project and don't know the code base well. Since this is quite a large change and I would really have it upstream rather than maintaining my own fork, how would you like to proceed? I propose that I have a look at your config flow and rewrite mine (you have implemented alternative 2, described in #1053). I don't know what a tag device is (I just started out with OCPP when I got my car a month ago), so not sure how to handle this. Once the rough implementation is ready, I would appreciate a review before I spend effort on tests and documentation updates. When that draft is acceptable, then I would proceed and update tests and documentation. Timeline is at least a couple of weeks (I prio this on my hobby list before it gets warm outside :) Would that work for you?

lbbrhzn:
Your proposal sounds fine to me! I really appreciate your help!

A tag is and RFID card or fob, that get scanned by the charge point. It can be used to start / stop transactions.
The integration currently only supports specifying tags in yaml. On my branch I added functionality to allow tags to be added/edited through the home assistant GUI.

My general strategy with the config / option flows was to allow as much parameters as possible to be changed through the option flow. As a result, config flows and options flows are nearly the same.  
It would be nice to allow a charge point entities to be automatically created when it connects for the first time. Same could be done for tags. This will save the user some time entering long tag numbers.

munterkalmsteiner:
> Your proposal sounds fine to me! I really appreciate your help!

Thanks!

> 
> A tag is and RFID card or fob, that get scanned by the charge point. It can be used to start / stop transactions. The integration currently only supports specifying tags in yaml. On my branch I added functionality to allow tags to be added/edited through the home assistant GUI.

I see. 

> 
> My general strategy with the config / option flows was to allow as much parameters as possible to be changed through the option flow. As a result, config flows and options flows are nearly the same. 

Ok.

> It would be nice to allow a charge point entities to be automatically created when it connects for the first time. Same could be done for tags. This will save the user some time entering long tag numbers.

Hm, I'm not sure I'm following you here. Isn't this already the case for charge points, i.e. a user creates a charge point device and when the charge point connects to the central system, the entities for the charge point are created, based on the configured/detected measurands. Or do you mean the charge point ***device*** is created as soon as it connects to the central system? 

Since I don't have any tags, I know nothing about how they work in the context of OCPP. How would they be discovered? Is this something that can be queried from the charge point?

drc38:
I don't use tags either but suggest it may be helpful to simply use HA's existing tags rather than have our own option flow for them. https://www.home-assistant.io/blog/2020/09/15/home-assistant-tags/ potentially it would allow authorisation by presenting a tag to your mobile (if it supports NFC) for example, and there may be other automation use cases where the charger tag can be used for other purposes in HA.

lbbrhzn:
some tests are still failing...

sandersr:
What else needs doing before this code can be merged?

munterkalmsteiner:
This PR is abandoned as I'm continuing the work by @lbbrhzn on his feature branch, see [here](https://github.com/munterkalmsteiner/ocpp/tree/new_option_flows_updated). But it is not finished yet and I don't see to have any time to work on it until the June/July.

lbbrhzn:
closing pull rerquest, since abandonded

Issue #1162, opened 2024-05-15
Integration could not load, after update to Homea Assistant Core 2024.5.3


juzel69:
After update to Version 2024.5.3 the integration could not be started. I removed the integration and reinstall it. But now it is not possible to start it.
![OCPP_Error](https://github.com/lbbrhzn/ocpp/assets/38318706/fef51c51-d00c-4228-9604-f4e9066320b4)

Regards Juerg

thomas6779:
Same here.

lutfijw:
It is because of ha.UnitOfFrequency.HERTZ declaration in const.py file. HA seem to removed it so the integration can not look up the value. Suggest to remove this. I did it and it work good.

thomas6779:
> It is because of ha.UnitOfFrequency.HERTZ declaration in const.py file. HA seem to removed it so the integration can not look up the value. Suggest to remove this. I did it and it work good.

thanks. That works fine!

JakobEgPedersen:
> It is because of ha.UnitOfFrequency.HERTZ declaration in const.py file. HA seem to removed it so the integration can not look up the value. Suggest to remove this. I did it and it work good.

Thank you, this solved my issue. It is working again now.

PR #1165, opened 2024-05-16
Remove ocpp hertz unit

### LINKED ISSUE #1161: ha.UnitOfFrequency.HERTZ error
after updating to the latest HomeAssistant stable version 2024.5.3, OCPP integration is failing due to unknown unit in const.py file.
I need to delete the ha.UnitOfFrequency.HERTZ due to it caused and error. Seem like Home Assistant is deleting the HERTZ unit.

drc38:
Fixes #1161 #1162


lbbrhzn (approved):
strange that this was removed. I recall there was a measurand that used it.

PR #1178, opened 2024-05-17
Update dep versions

drc38:
@lbbrhzn , I should have updated these in the deprecation fix!


lbbrhzn (approved):

lbbrhzn:
Oops..  i'll create a new release. Seems i'm making a mess off the releases today already!

PR #1242, opened 2024-07-08
improve reload

lbbrhzn:


drc38:
Hi @lbbrhzn,

Try changing:
`assert await async_setup_entry(hass, config_entry)`

with this:
`hass.config_entries.async_setup(config_entry.entry_id)`

lbbrhzn:
I think I should use config_entries.async_remove as well.

PR #1278, opened 2024-08-19
Workaround missing charge rate unit type in some chargers

malaakso:
At least Teison chargers return unknown key when getting supported charge rate unit types. In this case default to ampere.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,6 +1,6 @@
 {
     "name": "lbbrhzn/ocpp",
-    "image": "mcr.microsoft.com/devcontainers/python:3.11-bullseye",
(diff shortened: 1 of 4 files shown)


drc38:
Hi @malaakso, can you post your log showing the set charging profile messages of the code before your change and after please.

drc38:
Thanks, in that case I suggest changing the logic to test `if resp is None:` and then setting `resp = ChargingRateUnitType.amps.value` plus a log message.

The rest of the code does not then require amending. Also tests may need amending see #1283 for example.

malaakso:
I believe you mean `om.current.value` (="Current") instead of `ChargingRateUnitType.amps.value` (="A"). Please take a look now.

drc38 (approved):

Issue #1275, opened 2024-08-16
ABB Terra AC-W22-T-0 Bootloop


pvprodk:
**Describe the bug**
ABB Terra AC-W22-T-0 (listed as supported) keeps booting with these notifications showing up.
Controls are avaliable for a few seconds, until it reboots again.

**Screenshots**
![image](https://github.com/user-attachments/assets/b831ddf1-6763-44bd-a79d-f6d4dd465a66)

**Versions etc.**
OCPP v0.5.8
HA 2024.8.1
Charger firmware: V1.8.21 (Newest)

**Debug logging from HA (Not really sure if I'm including enough?)**

Traceback (most recent call last):
File "/usr/local/lib/python3.12/site-packages/websockets/legacy/protocol.py", line 963, in transfer_data
message = await self.read_message()
^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/site-packages/websockets/legacy/protocol.py", line 1033, in read_message
frame = await self.read_data_frame(max_size=self.max_size)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/site-packages/websockets/legacy/protocol.py", line 1108, in read_data_frame
frame = await self.read_frame(max_size)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/site-packages/websockets/legacy/protocol.py", line 1165, in read_frame
frame = await Frame.read(
^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/site-packages/websockets/legacy/framing.py", line 68, in read
data = await reader(2)
^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/asyncio/streams.py", line 752, in readexactly
await self._wait_for_data('readexactly')
File "/usr/local/lib/python3.12/asyncio/streams.py", line 545, in _wait_for_data
await self._waiter
File "/usr/local/lib/python3.12/asyncio/selector_events.py", line 1003, in _read_ready__data_received
data = self._sock.recv(self.max_size)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ConnectionResetError: [Errno 104] Connection reset by peer
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
File "/config/custom_components/ocpp/api.py", line 460, in post_connect
[...]

Issue #1263, opened 2024-07-30
Why i get those notifications?


une18:
![Screenshot_20240730_191840_Home Assistant](https://github.com/user-attachments/assets/0b17ad92-9300-4364-9442-8d5bfea1061f)
![Screenshot_20240730_191845_Home Assistant](https://github.com/user-attachments/assets/247b3037-9931-4f16-a49b-fa8d80cdb319)

Why I get these? It has always been 1-3 Time notes, now 136 pcs.

Issue #1126, opened 2024-04-09
Status intermittent between not available and available


StefanoGiu:
**Describe the bug**
Status intermittent between not available and available

**To Reproduce**
check the status of the integration

**Expected behavior**
Status should be stable and not become available every minute

**Desktop (please complete the following information):**
 - Version [e.g. 22] 0.5.2

StefanoGiu:
Any update? This custom component became unusable...

StefanoGiu:
Still happening.... any willingness or plan to fix this?

Aure51800:
Hi there, same here.
I can't find how to reproduce it. 
I know the last working version was 0.5.1
Entities become unavailable every 5s
What should I give you to debug ?

Aure51800:
Hi, after some testes, it happens every 40s, with errors:
Enregistreur: custom_components.ocpp
Source: custom_components/ocpp/api.py:223
intégration: Open Charge Point Protocol (OCPP) ([documentation](https://github.com/lbbrhzn/ocpp/blob/main/README.md), [problèmes](https://github.com/lbbrhzn/ocpp/issues))
S'est produit pour la première fois: 21:46:38 (12 occurrences)
Dernier enregistrement: 21:56:33

Skipping websocket subprotocol validation


And 

Enregistreur: custom_components.ocpp
Source: custom_components/ocpp/api.py:1020
intégration: Open Charge Point Protocol (OCPP) ([documentation](https://github.com/lbbrhzn/ocpp/blob/main/README.md), [problèmes](https://github.com/lbbrhzn/ocpp/issues))
S'est produit pour la première fois: 21:46:54 (10 occurrences)
Dernier enregistrement: 21:56:03

Unexpected exception in connection to 'TACW2244421S5041': 'list index out of range'
Traceback (most recent call last):
  File "/config/custom_components/ocpp/api.py", line 1020, in run
    await asyncio.gather(*self.tasks)
  File "/config/custom_components/ocpp/api.py", line 489, in post_connect
    resp = await self.get_configuration(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/ocpp/api.py", line 889, in get_configuration
    value = resp.configuration_key[0][om.value.value]
            ~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range

Aure51800:
And some more debugging logs

[...]

PR #1290, opened 2024-08-26
add try except for getting measurands

drc38:
fix #1275

PR #1292, opened 2024-08-28
Allow to disable autodetection of measurands

rinigus:
This PR is to address issues induced by ABB Terra AC firmware bug:
- #1275 
- probably #1263 
- probably #1126

As such bugs could be in other chargers, the solution is generic:

- Add config option to detect measurands automatically, set default to true

- Restrict checked measurands to the ones provided by users

I still need to add a quirk to make ABB chargers working again and update docs. So, it is WIP at this stage.

Todo list:

- [x] Notify ABB regarding the issue
- [x] Add a configuration option to disable automatic detection of the measurands. This will be permanently enabled for ABB chargers and others that could have similar issues.
- [x] Add a quirk to set MeterValuesSampledData to the one specified in configuration by users.
- [x] Use form to select measurands if not autodetected
- [x] Update documentation on how to configure ABB Terra AC charger.

**Edit:** PR is ready for review

Fixes: #1275 
Fixes: #1263 
Fixes: #1126


drc38:
Hi @rinigus, I'm curious as to why the boolean "quirk" setting is needed when you already have the disable autodetection setting?

If autodetection is disabled a user supplied list of measurands must be used, would it not?

rinigus:
Quirk is needed to set MeterValuesSampledData back to some reasonable value on the charger. Without it, every time OCPP server calls GetConfiguration(MeterValuesSampledData), we get charger reboot. That is if users earlier used automatic config for measurands and, due to ABB charger bugs, OCPP server set MeterValuesSampledData to the measurands list with all possible measurands.

I expect that all ABB charger users that tested current OCPP HA intergration have all that setting in such state and cannot use HA due to it.  So, until ABB will fix the firmware, we have to have that quirk. Later, it should be possible to remove it, unless some other charger with such issue comes along.

rinigus:
This PR is ready for review. Few decisions that were made while developing it:

1. I have restored older form for selecting the measurands. This was done to keep nice user interface mainly. Originally, I kept the string with the list of measurands and verified it on submission, but that was problematic as when user entered the wrong measurand it was returned back to the original form with all data reset. I couldn't figure out how to add custom verification in the form, as it mainly struggled with the serialization used in https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/config_validation.py. So, in the end, it looked the best way to ensure that the user can pick and choose only correct values.
2. The measurand form has now every line starting with OCPP tag. It was added in all supported languages. Its done to simplify writing docs on which options to choose and allow users to check for available tags in charger firmware docs. For example, ABB docs list only tags, not their meaning. 
3. I have reverted try/except approach for GetConfiguration from #1290 in apy.py. If we want to use this construction then it probably has to be sprinkled through out the code for other similar cases. However, I would suggest not to use try/except unless it is necessary. Namely, as it is, it just masked the problem and didn't resolve it. In this partictlar case, it kept crashing just a bit later (see last log in https://github.com/lbbrhzn/ocpp/issues/1275#issuecomment-2313314865).

Other than that, I have tried to stick to the code style. The fix was applied to my charger and it looks like it helped to avoid the issue.  Documentation has been updated to help users to fix their chargers as well. 


Re contact with ABB: no reply from them yet.

lbbrhzn:
Great, I'll take a look in the next coming days. BTW it would also be nice to have the measurable selectable in an optionsflow. 
I experimented with that once, with some care you can reuse the one for the configflow

rinigus:
Thank you very much! Let's get this over the line first and then would be nice to make configflow working, indeed. Just would have to learn it...

custom_components/ocpp/api.py
```diff
@@ -464,30 +467,54 @@ async def handle_set_charge_rate(call):
...
-            accepted_measurands = ",".join(accepted_measurands)
+                accepted_measurands = ",".join(accepted_measurands)
+            else:
+                accepted_measurands = all_measurands
+                _LOGGER.debug(
+                    f"'{self.id}' measurands set manually to {accepted_measurands}"
+                )
 
             if len(accepted_measurands) > 0:
                 _LOGGER.debug(
```
drc38, line 520: I suggest replacing the `await self.configure(key, accepted_measurands)` here with
```
req = call.ChangeConfiguration(key=key, value=all_measurands)
resp = await self.call(req)
```

custom_components/ocpp/api.py
```diff
@@ -464,30 +467,54 @@ async def handle_set_charge_rate(call):
...
+                    f"'{self.id}' quirk: disable setting OCPP measurands via the quirk if your charger does not need it anymore"
                 )
-                chgr_measurands = all_measurands
 
-            accepted_measurands = []
-            cfg_ok = [
-                ConfigurationStatus.accepted,
-                ConfigurationStatus.reboot_required,
-            ]
+            chgr_measurands = await self.get_configuration(key)
```
drc38, line 495: If the try/except is kept here then the quirk code above and config setting is no longer required. Noting also my comment below to avoid the get configuration being called prior to setting the measurands.

rinigus: @drc38, interesting suggestion. So, together with the other one on using `call.ChangeConfiguration` directly, we would have:

1. try/except protected `self.get_configuration`
2. forcing configuration by `call.ChangeConfiguration` depending on try/except and other settings

Problem is that it doesn't work on the charger side if the charger configuration is already faulty. During `self.get_configuration` the chargers rebooting that actually leads to new boot notification of the charger to OCPP server and we start calling `post_connect` . And the cycle repeats. That case is reflected in the last log of https://github.com/lbbrhzn/ocpp/issues/1275#issuecomment-2313314865

So, unfortunately, we need a quirk for these chargers.

As the quirk is still needed, I would suggest to keep `self.configuration` as well. Namely, at some point, we maybe can get rid of the quirk (ABB updates firmware, users have chargers "repaired", ...). Then keeping the same way of setting configuration as for other parameters would keep the code consistent.

What do you think about it?

rinigus:
I have an issue with linting when committing api.py. On pre-commit checks, `isort` pushes imports as they are in my version. By the way, they are already like that in the main repo. In principle, it seems that the suggested ordering in your Github linting check is better (to my taste), but, with the pre-commit hooks configured as they are, impossible to do.

How to proceed with it?

custom_components/ocpp/api.py
drc38: Ah ok, I hadn't realised the initial get configuration call caused a charger reboot.

drc38: You could put `chgr_measurands = await self.get_configuration(key)` inside the `if autodetect_measurands:` statement and include the call change configuration in the else statement. It would result in setting the measurands twice during first connection, but would avoid having to remove the quirk code later on.

rinigus: Sure, we can do it that way. So, to avoid quirk, we can also move chrg_measurands to later stage as in:

```python
            if autodetect_measurands:
                accepted_measurands = []
                cfg_ok = [
                    ConfigurationStatus.accepted,
                    ConfigurationStatus.reboot_required,
                ]

                for measurand in all_measurands.split(","):
                    _LOGGER.debug(f"'{self.id}' trying measurand: '{measurand}'")
                    req = call.ChangeConfiguration(key=key, value=measurand)
                    resp = await self.call(req)
                    if resp.status in cfg_ok:
                        _LOGGER.debug(f"'{self.id}' adding measurand: '{measurand}'")
                        accepted_measurands.append(measurand)

                accepted_measurands = ",".join(accepted_measurands)
            else:
                accepted_measurands = all_measurands
                req = call.ChangeConfiguration(key=key, value=all_measurands)
                resp = await self.call(req)
                _LOGGER.debug(
                    f"'{self.id}' measurands set manually to {accepted_measurands}"
                )

            chgr_measurands = await self.get_configuration(key)
```

Such combination should work if the user has selected correct measurands. The only problem is that we would "normalize" quirky behavior. Obviously, it would have to be explained in the code why it is in such manner and maybe that is sufficient. 

With the code above, quirk is not needed and can be removed.

If you think it is better, I would make the changes in the code (config, tests) and docs. But would like to be sure that this is what is expected and prefer not to have later to change it back :)

drc38: Let's see what @lbbrhzn prefers

drc38:
> I have an issue with linting when committing api.py. On pre-commit checks, `isort` pushes imports as they are in my version. By the way, they are already like that in the main repo. In principle, it seems that the suggested ordering in your Github linting check is better (to my taste), but, with the pre-commit hooks configured as they are, impossible to do.
> 
> How to proceed with it?

It is probably worth migrating to ruff for both @lbbrhzn?

lbbrhzn:
Yes, we should migrate to ruff, both for the plugin used by the editor and for the the precommit.

rinigus:
@lbbrhzn : OK, then I will wait for your review before making changes in PR if needed

custom_components/ocpp/api.py
lbbrhzn: I prefer teh code with no quirk

rinigus: OK, then I will rewrite that bit and change the docs accordingly.

lbbrhzn:
I updated the main branch to use ruff, and merged the changes into this branch.

rinigus:
@lbbrhzn I was going to rebase this branch, but merge would work as well. Will resume working on this in few days as I am in a middle of implementing one feature in unrelated project. So, it makes sense to finish it before jumping back to OCPP.

rinigus:
@drc38 and @lbbrhzn: as suggested, I have dropped the quirk. For all chargers that request manual measurand configuration, we bypass checking current state of `MeterValuesSampledData` and set it to user provided list. This allows all ABB users now to get their chargers back and working.

Please review

lbbrhzn:
Ready to merge to main?

rinigus:
Should be. At least it works with my charger :) . I did try not to change anything in the autodetection part of the code - so it should work for earlier chargers as well. Would be good to review that bit just in case, but on my side, it should be ready.

lbbrhzn (approved):
looks good to me! Let's rock!

CRDAU:
Just want to say thanks! I updated firmware / occp combo yesterday after a long time of no updates and this drove me mad until i found this thread and worked through applying the workaround.

Only question I have is -> is there any easy way to 'check' through the other measureands (and identify the dodgy ones?) 
My guess is 'no' unless you delete / reinstall the integration every time because you can only access that initial configuration screen on setup. (i'm not quite adept yet at playing with the actual code...)
There's a couple I found useful that I'd like to still have access too - but too scared to break things again now that I've got my charger stable again.

rinigus:
> Only question I have is -> is there any easy way to 'check' through the other measureands (and identify the dodgy ones?)

best is to check docs for the charger. we would have to implement simpler system for changing configuration, its just not done at this stage. as for not really liking to play with the code you don't understand - was just in the same boat with you. this is the first PR that I did for HA. it has a learning curve, but you could start from reading HA developer docs and the code. as far as I understand, we need to implement https://developers.home-assistant.io/docs/config_entries_options_flow_handler for simpler config changes

CRDAU:
Ah okay thanks! So the odd thing is... I've noticed after a couple of weeks of running with the fix Etc... Those other measurands/entities I'm interested in, are actually updating!! And nothing appears to be broken. 
So I don't know what's going on there but I'm not going to complain

PR #1313, opened 2024-09-08
Set branch coverage to false

drc38:
@lbbrhzn I don't think branch testing is working properly with the asyncio tests

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,5 @@
 [run]
-branch = True
+branch = False
 source = custom_components/ocpp
 
 [report]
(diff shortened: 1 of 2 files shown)


lbbrhzn:
Why disable branch coverage? It would mean you only get coverage info after merging to main. I think we should look at coverage before merging to main.

drc38:
> Why disable branch coverage? It would mean you only get coverage info after merging to main. I think we should look at coverage before merging to main.

It does not relate to branches in Github but rather branching code eg an if statement. See https://coverage.readthedocs.io/en/latest/branch.html#branch

lbbrhzn:
ok, then it's fine.

lbbrhzn (approved):

Issue #1449, opened 2024-12-27
Charger power reactive export/import have lost their state classes and units after update to v0.6.3


hummermann87:
**Describe the bug**
Charger power reactive export/import have lost their state classes and units after update to v0.6.3.

**To Reproduce**
N/A

**Expected behavior**
No errors

**Screenshots**
![image](https://github.com/user-attachments/assets/5b0a3b92-3d0d-4f4a-a977-8ababd8cca4d)

**Desktop (please complete the following information):**
 - OS: Home Assistant, OS 14.1, Core 2024.12.5, Supervisor 2024.12.0

**Additional context**

N3rdix:
Good catch, I missed that when introducing the reactive power device_class. I'll fix it asap

BitsLeftOver:
Same problem.

N3rdix:
Should be fixed once #1451 is included in an upcomibg Release

PR #1417, opened 2024-11-30
Correct reactive power device class

N3rdix:
fixes #1403 partially by giving a device class for reactive power sensors.

There is no device class for reactive energy so far, means this cannot be fixed and does not have a device class assigned on purpose: https://developers.home-assistant.io/docs/core/entity/sensor/
Once HA provides a device class for it it can be added here...

PR #1451, opened 2024-12-28
Fix missing state class for reactive power sensors after #1417

N3rdix:
After explicitly switching to reactive power device class I missed the state_class also depending on this device_class (#1417).

Should fix https://github.com/lbbrhzn/ocpp/issues/1449 (the raised repair regarding the changed unit of measurement is expected nevertheless)

### DIFF
--- a/tests/test_charge_point_v16.py
+++ b/tests/test_charge_point_v16.py
@@ -268,6 +268,8 @@ async def test_cms_responses_v16(hass, socket_enabled):
     assert int(cs.get_metric("test_cpid", "Current.Import")) == 0
     assert int(cs.get_metric("test_cpid", "Voltage")) == 228
     assert cs.get_unit("test_cpid", "Energy.Active.Import.Register") == "kWh"
+    assert cs.get_ha_unit("test_cpid", "Power.Reactive.Import") == "var"
+    assert cs.get_unit("test_cpid", "Power.Reactive.Import") == "var"
     assert cs.get_metric("unknown_cpid", "Energy.Active.Import.Register") is None
     assert cs.get_unit("unknown_cpid", "Energy.Active.Import.Register") is None
     assert cs.get_extra_attr("unknown_cpid", "Energy.Active.Import.Register") is None
@@ -821,7 +823,7 @@ class ChargePoint(cpclass):
                             "value": "89.00",
                             "context": "Sample.Periodic",
                             "measurand": "Power.Reactive.Import",
-                            "unit": "W",
+                            "unit": "var",
                         },
                         {
                             "value": "0.010",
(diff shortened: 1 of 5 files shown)


drc38:
@N3rdix it'd be useful to include an assert in the tests to confirm the unit returned for this device class is what is expected.

N3rdix:
@drc38 although I added an assert in the test_init.py it does not feel to be the best place. As we do not have a test_sensor.py yet and I am not confident enough yet with pytest to build such a test from scratch easily I thought an experienced view might help.
Happy for any kind of help 😄

drc38:
> @drc38 although I added an assert in the test_init.py it does not feel to be the best place. As we do not have a test_sensor.py yet and I am not confident enough yet with pytest to build such a test from scratch easily I thought an experienced view might help. Happy for any kind of help 😄

Best to put the assert statements here https://github.com/lbbrhzn/ocpp/blob/d351da0c7cc7f4a0dcc135e6ddffe2d9a5814b1b/tests/test_charge_point_v16.py#L270
Using mock config creates the integration's sensors so no need to do so manually. Hope that helps.

drc38:
Also add the device class here https://github.com/lbbrhzn/ocpp/blob/d351da0c7cc7f4a0dcc135e6ddffe2d9a5814b1b/custom_components/ocpp/const.py#L127

N3rdix:
Hi @drc38,
thanks for your hints, that helped to understand the tests better 👍 

I've added the asserts and adjusted the mock data accordingly. Do you think this is enough/ok?

tests/test_charge_point_v16.py
```diff
@@ -821,7 +823,7 @@ async def send_meter_periodic_data(self):
                             "value": "89.00",
                             "context": "Sample.Periodic",
                             "measurand": "Power.Reactive.Import",
-                            "unit": "W",
+                            "unit": "var",
```
drc38, line 826: @N3rdix looks good to me, you could remove this unit line which would confirm the default unit of var is being assigned by the integration.

drc38 (approved):

tests/test_charge_point_v16.py
N3rdix: When I do that the asserts both fail but I could not yet find the reason yet (must be test related as the integration itself creates the HA entity correctly now in the devcontainer) 😕

drc38: Ahhh, that's because the metric class is not actually the sensor but a representation of the raw data read from the charger. To get the sensor value or unit for the assert it'd be similar to line 14 in this https://github.com/MatthewFlamm/pytest-homeassistant-custom-component/blob/master/tests/test_sensor.py

N3rdix: That added the missing piece to my head 😄 I've added a separate `test_sensor.py`, I think this might be helpful for future sensor tests anyways

drc38 (approved):

tests/test_charge_point_v16.py
drc38: Just be aware that in order to test/assert in a sensor how data is processed you'd need to set the metric value/unit in the central system first as there is no dummy charger injecting data.

lbbrhzn (approved):

PR #1469, opened 2025-01-05
Create update-version.yml

lbbrhzn:
add github action to update version field in manifest.json on publish


drc38:
And remove https://github.com/lbbrhzn/ocpp/blob/main/manage/update_manifest.py?

PR #1710, opened 2025-09-08
Average Current Values. More robust handling of meter values.

### LINKED ISSUE #1709: After updating to 0.10 Growatt Thor 11A/S: Session Energy and Active Import Register sensors are always 0
This issue is the same as #1466 but re-introduced in version 0.10.0
It might be related to #1707 

In 0.10.0 the code logic for handling DEFAULT_MEASURAND/Energy.Active.Import.Register and session energy was changed and for an unknown reason disabled for OCPP v1.6. 

I tried to re-enable the previous logic locally but it does not work as expected for some reason, even when reverting to 0.9.0 style code

jthun:
Average Current values. Handle cases where meter values bucket contains several EAIR with different contexts (Fixes #1709). Set energy_session initally to 0 to avoid spike. More tests.

## Summary by CodeRabbit

- New Features
  - Multi-connector aware meter handling with priority-based energy selection and improved session energy calculations.
  - Smarter measurand configuration with autodetect and fallbacks.
  - Backstop and staged status triggering for non-compliant chargers.

- Bug Fixes
  - Safer connector count handling and robust processing of missing/invalid measurements.
  - Timeout-bounded boot/status notifications; cleaner failure handling.
  - Reduced noise on charge rate errors.

- Refactor
  - Per-phase aggregation and clearer energy routing for voltage/current/power metrics.

- Tests
  - Extensive new async test suites and fixtures covering configuration, transactions, firmware, and resilience.

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -290,10 +290,19 @@ class CentralSystem:
             await charge_point.reconnect(websocket)
 
     def _get_metrics(self, id: str):
-        """Return metrics."""
+        """Return (cp_id, metrics mapping, cp instance, safe int num_connectors)."""
         cp_id = self.cpids.get(id, id)
         cp = self.charge_points.get(cp_id)
-        n_connectors = getattr(cp, "num_connectors", 1) or 1
+
+        def _safe_int(value, default=1):
+            try:
+                iv = int(value)
+                return iv if iv > 0 else default
+            except Exception:
+                return default
+
+        n_connectors = _safe_int(getattr(cp, "num_connectors", 1), default=1)
+
         return (
             (cp_id, cp._metrics, cp, n_connectors)
             if cp is not None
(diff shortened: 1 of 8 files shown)


drc38 (approved):

jthun:
Is that using the code in the PR or 0.10.1? Are there other sensors (maybe inactive) on the charger device? Can you start a charging session and capture the logs a little longer? I just want to see that the charger is actually sending periodic EAIR updates.

drc38:
> Is that using the code in the PR or 0.10.1? Are there other sensors (maybe inactive) on the charger device? Can you start a charging session and capture the logs a little longer? I just want to see that the charger is actually sending periodic EAIR updates.

It's using this PR (although both are equivalent for this issue). Looking at the codebase it is ignoring when an EAIR is sent with a transactionId and connector=1, which are the Sample.Periodic values. For a single connector charger it should assume an EAIR associated with connector 1 is the same as the EAIR for connector 0 unless the value received is less than the previous EAIR (eg ABB chargers start the transaction EAIR at 0 for each session) ie not monotonic.

PR #1984, opened 2026-06-27
Fix WebSocket handshake failure for chargers missing Connection: Upgrade header

vavallee:
### Problem
Some charger firmware (e.g. **Autel MaxiCharger US AC LW10-N14**, fw v1.38.00/v1.22.00) omits the required `Connection: Upgrade` header from the WebSocket opening handshake. `websockets` 14+ strictly validates this header and raises `InvalidUpgrade`, so the affected chargers can **never connect** to the OCPP central system.

### Fix
Add a `process_request` hook (`_fix_missing_connection_header`) that injects `Connection: upgrade` before the library validates the handshake, only when the header is absent. Logs a warning recommending a firmware update. No effect on compliant chargers (covered by a test asserting no warning is logged when the header is present).

### Notes
- Rebased onto **v0.10.15**; `api.py` is otherwise unchanged in that range so this applies cleanly.
- Module-level function + unit tests in `tests/test_api_paths.py`.
- Supersedes #1933, which was **auto-closed by the stale bot** without a human review — not on technical grounds. Friendly ping @lbbrhzn for a look when you have a moment; happy to adjust.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket connection handling for chargers with firmware that omits the required upgrade header.
  * Connection attempts that previously failed during the initial handshake should now proceed more reliably.
  * Added safeguards to preserve valid connection headers when they are already present.

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -16,6 +16,7 @@ from homeassistant.exceptions import HomeAssistantError
 import homeassistant.helpers.config_validation as cv
 import voluptuous as vol
 from websockets import Subprotocol, NegotiationError
+from websockets.http11 import Request, Response
 import websockets.server
 from websockets.asyncio.server import ServerConnection
 
@@ -180,10 +181,33 @@ class CentralSystem:
         else:
             self.ssl_context = None
 
+        async def _fix_missing_connection_header(
+            connection: ServerConnection, request: Request
+        ) -> Response | None:
+            """Work around a bug in some charger firmware (e.g. Autel MaxiCharger
+            US AC LW10-N14, firmware v1.38.00/v1.22.00) that omits the required
+            Connection: Upgrade header from the WebSocket opening handshake.
+
+            websockets 14+ strictly validates this header and raises
+            InvalidUpgrade when it is absent, preventing the charger from
+            ever connecting.  We inject the header before the library validates
+            it so that the handshake can proceed normally.
+            """
+            if not request.headers.get_all("Connection"):
+                _LOGGER.warning(
+                    "Charger at %s sent a WebSocket upgrade request without the "
+                    "required 'Connection: Upgrade' header. Injecting header as "
+                    "a workaround — consider updating the charger firmware.",
+                    connection.remote_address,
+                )
+                request.headers["Connection"] = "upgrade"
+            return None
+
         server = await websockets.serve(
             self.on_connect,
             self.settings.host,
             self.settings.port,
+            process_request=_fix_missing_connection_header,
             select_subprotocol=self.select_subprotocol,
             subprotocols=self.subprotocols,
             ping_interval=None,  # ping interval is not used here, because we send pings mamually in ChargePoint.monitor_connection()
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -38,6 +38,29 @@ from .chargepoint import SetVariableResult
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 # Uncomment these when Debugging
 # logging.getLogger("asyncio").setLevel(logging.DEBUG)
+
+
+async def _fix_missing_connection_header(
+    connection: ServerConnection, request: Request
+) -> Response | None:
+    """Work around a bug in some charger firmware (e.g. Autel MaxiCharger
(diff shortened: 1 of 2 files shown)


proscar87:
**Independent reproduction & verification** (not a review request — just sharing evidence in case it's useful when this gets looked at).

I reproduced the bug and verified the fix in an isolated environment (no Autel hardware needed): fresh clone of `lbbrhzn/ocpp` at `main` (`33721c0`), Python 3.14.3, `websockets==16.1.1` (satisfies `manifest.json`'s `websockets>=14.1`) installed via `uv pip install -r requirements.txt`. I built a minimal server using the exact same `websockets.serve(...)` call shape as `CentralSystem.create()` (`select_subprotocol`, `subprotocols=["ocpp1.6"]`, and, for the "patched" run, `process_request=_fix_missing_connection_header`), then drove handshakes over a raw TCP socket so I could omit headers exactly like the affected firmware does — a real `websockets`/browser client won't let you send a request missing `Connection`.

**1. Reproduced the bug on unpatched `main`**

Handshake with `Upgrade: websocket` but no `Connection` header, mirroring the firmware behavior described in this PR:

```
GET / HTTP/1.1
Host: 127.0.0.1:9411
Upgrade: websocket
Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==
Sec-WebSocket-Version: 13
Sec-WebSocket-Protocol: ocpp1.6
```

Server raises exactly the exception described in the PR:
```
websockets.exceptions.InvalidUpgrade: missing Connection header
```
and the client receives `HTTP/1.1 426 Upgrade Required` — the charger never connects.

**2. Verified the fix resolves it**

Same raw handshake, same server, this time with this PR's diff applied on top of current `main` (`process_request=_fix_missing_connection_header` wired into `websockets.serve`):
- Server logs the expected warning: `Charger at (...) sent a WebSocket upgrade request without the required 'Connection: Upgrade' header. Injecting header as a workaround...`
- Client receives `HTTP/1.1 101 Switching Protocols` — connection succeeds.

**3. Confirmed the normal (compliant) case is unaffected**

Sent a well-formed handshake (with `Connection: Upgrade` present) against both the unpatched and patched server: `101 Switching Protocols` in both, no warning logged in either case. Compliant chargers see no behavior change.

**4. Test suite**

Ran the repo's suite the same way CI does (`pytest --timeout=60 -n auto -q tests`):
- `main` unpatched: **169 passed, 0 failed**, 95% coverage.
- this PR's diff applied on top of current `main`: **171 passed** (169 + the 2 new tests), **0 failed**, 95% coverage.
- `ruff check` on the two changed files: clean. `ruff format --check` flags one pre-existing formatting nit in `tests/test_api_paths.py` (an `assert all(...)` a few hundred lines above the new tests, unrelated to this diff) — that drift is already on `main` today, not introduced by this PR.

**5. Risk analysis**

Wanted to check whether the workaround relaxes validation more broadly than necessary:
- The hook only fires when `Connection` is *completely absent* (`not request.headers.get_all("Connection")`). I also tried a *present-but-wrong* header (`Connection: keep-alive`) and an empty-string value — both are still correctly rejected downstream by `websockets`' own `ServerProtocol.accept()` (`InvalidUpgrade` / `InvalidHeaderFormat`). The patch doesn't touch malformed-but-present headers, only the "missing entirely" case.
- Ordering is correct and matters here: reading `websockets/asyncio/server.py`, the user-supplied `process_request` hook runs *before* `ServerProtocol.accept()`'s strict RFC 6455 validation, so `Upgrade`, `Sec-WebSocket-Key`, `Sec-WebSocket-Version` and subprotocol negotiation are all still fully enforced afterward. This patch touches nothing but the one missing header.
- Minor nit, not a blocker: the log message says `Connection: Upgrade` but the injected value is lowercase `"upgrade"`. Harmless (`websockets` accepts it case-insensitively) — just noting it in case it's worth matching case in the log string for readability.

I didn't find a scenario where this accepts a handshake it shouldn't, and it fixes exactly the failure mode described (confirmed against the exact `InvalidUpgrade: missing Connection header` exception from `websockets` 14+). Happy to try other edge cases (duplicate `Connection` headers, HTTP/1.0 clients, etc.) if that would help move this along.

drc38:
not intending to merge, as the charger firmware needs to address non-compliance with the websocket standard.

PR #2000, opened 2026-07-22
Add retry logic with exponential backoff to SetChargingProfile

### LINKED ISSUE #1999: SetChargingProfile should retry on transient timeout failures
## Problem

When a \`SetChargingProfile\` OCPP call times out (e.g., due to temporary network lag or charger processing delay), the function catches the exception, logs a warning, and sends a persistent notification to the HA UI. However, it returns \`False\` without retrying, and the charging rate update is lost.

**Impact:** Dynamic charge rate adjustments (e.g., load-balancing based on available grid current) fail on any transient network hiccup, even brief ones. A single 10-second timeout causes the operation to give up, requiring manual intervention to retry.

**Root cause:** Transient network failures are treated as permanent failures. A brief lag or charger processing delay causes immediate failure instead of tolerating the delay with a retry.

## Solution Approach

Add retry logic with exponential backoff for transient failures (timeouts/exceptions), while maintaining immediate failure for active charger rejections (which are policy decisions, not glitches).

## Environment

- Integration version: v0.10.15
- OCPP Version: 1.6
- Scenario: Dynamic load-balancing with frequent charger target current adjustments during active charging sessions

Mark-Hetherington:
Fixes #1999

When SetChargingProfile times out due to transient network issues or charger processing delay, retry up to 3 times with exponential backoff (0.5s, 1s) before notifying the user. Immediate charger rejections still fail fast.

This prevents dynamic charge rate adjustments from silently failing on brief network hiccups.

## Changes
- Adds asyncio import for sleep() during retry backoff
- Retries up to 3 times for exceptions (transient failures)
- Fails immediately for active rejections (charger policy)
- Only notifies user if all retries exhausted
- Includes debug logging for retry attempts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when applying charging profiles by retrying failed requests up to three times.
  * Added short backoff delays between attempts to recover from temporary communication issues.
  * Charging requests now stop immediately on acceptance or explicit rejection, while repeated exceptions still trigger warning notification after the final attempt.

### DIFF
--- a/custom_components/ocpp/ocppv16.py
+++ b/custom_components/ocpp/ocppv16.py
@@ -1,5 +1,6 @@
 """Representation of a OCPP 1.6 charging station."""
 
+import asyncio
 from datetime import datetime, timedelta, UTC
 import logging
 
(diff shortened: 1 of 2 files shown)


Mark-Hetherington:
@coderabbitai review

drc38:
A more generalised approach would be preferred ie if a command timesout place it in a queue, fail the websocket connection and then on re-establishing the websocket connection process the queue in fifo order.

Mark-Hetherington:
Closing this PR. redesigning the timeout handling from ad-hoc per-call retries to a general mechanism.

PR #2012, opened 2026-07-28
Fix OCPP subprotocol negotiation and version handling across reconnects

### LINKED ISSUE #2008: OCPP 2.0.1 charge point rejected: 2.0.1 BootNotification validated against OCPP 1.6 schema (FormatViolationError)
## Description

When a charge point is switched from OCPP 1.6J to **OCPP 2.0.1**, it can no longer establish a session with the integration. The charger connects and sends what appears to be a spec-compliant OCPP 2.0.1 `BootNotification` (top-level members `chargingStation` and `reason`), but the payload is validated against the **OCPP 1.6** `BootNotification` schema. This raises a `jsonschema` `ValidationError` ("Additional properties are not allowed ('chargingStation', 'reason' were unexpected)"), wrapped as `ocpp.exceptions.FormatViolationError`, and the connection is dropped. The charge point never comes online and all its entities stay `unavailable`/stale.

Switching the same charger **back to OCPP 1.6J immediately restores full operation** (control, status, meter values, GetConfiguration, etc.).

`chargingStation` and `reason` are exactly the top-level members of the **OCPP 2.0.1** `BootNotificationRequest`, so the incoming 2.0.1 message appears to be routed through the v1.6 `ChargePoint`/message set rather than the v2.0.1 one — i.e. the negotiated subprotocol/version isn't selecting the 2.0.1 schemas.

## Environment

- **Integration:** lbbrhzn/ocpp **v0.10.15**
- **Library requirement (manifest):** `ocpp>=2.1.0`, `websockets>=14.1`
- **Home Assistant:** 2026.7.4 (HAOS), **Python 3.14**
- **Charge point:** FoxESS A-Series **A022KP1**, firmware **1.10 / 1.06**, connector `Available`, `SupportedFeatureProfiles = Core,FirmwareManagement,LocalAuthListManagement,Reservation,SmartCharging,RemoteTrigger`
- Charger works fully on **OCPP 1.6J**; fault appears only after switching the charger to **OCPP 2.0.1**.

## Steps to reproduce

1. A charge point working normally on OCPP 1.6J with lbbrhzn/ocpp v0.10.15.
2. Switch the charger's OCPP profile to **2.0.1** (charger-side setting).
3. Charger reconnects and sends its `BootNotification`.
4. Observe the `FormatViolationError` below; no session is established and the charge-point entities remain `unavailable`.

## Traceback (paths redacted)

```
File ".../custom_components/ocpp/chargepoint.py", line 534, in _handle_call
File ".../site-packages/ocpp/charge_point.py", line 300, in _handle_call
File ".../site-packages/ocpp/charge_point.py", line 272, in route_message
File ".../site-packages/ocpp/messages.py", line 179, in validate_payload
File ".../site-packages/ocpp/messages.py", line 240, in _validate_payload
File ".../site-packages/ocpp/messages.py", line 247, in _validate_payload
File ".../site-packages/jsonschema/validators.py", line 450, in validate
jsonschema.exceptions.ValidationError: Additional properties are not allowed ('chargingStation', 'reason' were unexpected)

    raise FormatViolationError(
ocpp.exceptions.FormatViolationError: Payload for Action is syntactically incorrect or structure
for Action, {'cause': "Additional properties are not allowed ('chargingStation', 'reason' ...)"}
```

Loggers involved: `ocpp`, `custom_components.ocpp`, `homeassistant.components.websocket_api.http.connection`.

## Expected behaviour

A charge point negotiating the `ocpp2.0.1` subprotocol should be handled with the OCPP 2.0.1 message schemas, so a 2.0.1 `BootNotification` (`chargingStation` + `reason`) validates and the session is established.

## Actual behaviour

The 2.0.1 `BootNotification` is validated against the OCPP 1.6 schema → `ValidationError` → `FormatViolationError` → connection dropped; charge point never registers.

## Additional notes

- The config flow exposes **no OCPP-version / subprotocol option** (fields are: `host, port, ssl, ssl_certfile_path, ssl_keyfile_path, csid, websocket_close_timeout, websocket_ping_tries, websocket_ping_interval, websocket_ping_timeout`), and there is no options flow — so there is no user-facing way to influence version/subprotocol selection.
- Happy to capture the full unredacted traceback, a raw WebSocket capture of the BootNotification, or enable debug logging if useful.

KingHavok:
Fixes #2008

## Background

Setting up a FoxESS A-series charger (A022KP1) against this integration was
unreliable until the charger was locked to OCPP 1.6J on the charger side; switching
it to 2.0.1 (to explore ISO 15118/SoC support) produced the `FormatViolationError`
loop in #2008, and switching back to 1.6 did not recover — only a config-entry
reload did. To isolate whether the charger or the integration was at fault, the
charger was pointed at a standalone test CSMS and its handshakes captured in all
three of its modes (1.6J / 2.0.1 / auto). The charger turned out to be well-behaved
(it honours whatever subprotocol is negotiated and holds a stable 2.0.1 session);
the captures identified the integration-side root causes fixed here.

## Problem

A charge point that supports several OCPP versions can wedge the integration into a
`FormatViolationError` loop that only a config-entry reload clears. Root cause was
confirmed on the wire against a FoxESS A022KP1 using a standalone test CSMS:

- After an OCPP version switch, the charger makes a **short-lived `ocpp1.6` probe
  connection** (~3 s, self-dropped) before its real connection. The probe caused a
  v1.6 `ChargePoint` to be built and **cached** for the `cp_id`. The real 2.0.1
  connection then reused the cached v16 object, and its 2.0.1 `BootNotification`
  (`chargingStation`/`reason`) was validated against the **1.6 schema**:

  ```
  jsonschema.exceptions.ValidationError: Additional properties are not allowed
      ('chargingStation', 'reason' were unexpected)
  ocpp.exceptions.FormatViolationError: Payload for Action is syntactically incorrect ...
  ```

  This repeats on every reconnect (both directions: 1.6→2.0.1 and 2.0.1→1.6);
  charger power-cycles don't clear it, only reloading the config entry does.

- Independently, `select_subprotocol` iterated a `set()` of the charger's offer, so
  when a charger offered several subprotocols (this charger's "auto" mode offers
  `ocpp2.0.1, ocpp1.6`) the negotiated version depended on set-iteration order and
  could vary between handshakes.

## Changes

- **`on_connect`:** when a charger reconnects negotiating a *different* OCPP version
  than the cached `ChargePoint` was built with, stop the stale object and rebuild it
  from the current handshake's subprotocol instead of reusing it. Same-version
  reconnects keep the existing fast `reconnect()` path.
- **`select_subprotocol`:** iterate the server's ordered subprotocol list, which is
  the documented behaviour of the `websockets` default this function overrides
  (*"pick the first one in the list declared the server"*). The override closely
  mirrors that default but iterated the client's offer rather than the server's
  declared list, which is what made selection non-deterministic. The other deviation
  from the default — returning `None` instead of raising when the client offers no
  subprotocol — is deliberate and is **preserved**, since it is what allows a charger
  offering no subprotocol to default to OCPP 1.6. This restores the intended
  deterministic behaviour. Chargers offering a single subprotocol are unaffected; a
  dual-stack charger now consistently negotiates the first entry of
  `DEFAULT_SUBPROTOCOLS` (`ocpp1.6`) instead of an arbitrary one, so a user whose
  process happened to select 2.0.1 through the old unordered set may see the
  negotiated version change — the new OCPP-version option is the explicit lever for
  that.
- **New config-flow option "OCPP version"** (default `auto`, preserving current
  behaviour). Pinning a version restricts negotiation to that version's subprotocol,
  so mixed-version probes are rejected at the websocket handshake and never reach the
  charge point cache. When pinned to a 2.x version, connections offering **no**
  subprotocol are rejected too (previously they silently became v1.6 charge points).
- **Reconfigure flow** (`async_step_reconfigure`): existing entries can change
  central-system settings — including the new pin — without deleting and re-adding
  the integration.
- Debug log of the charger's offered subprotocols to aid diagnosis.

## Scope note

The pin applies to the whole central-system entry (one websocket port). Mixed-version
installations should keep `auto` (which now self-heals thanks to the rebuild) or use
separate ports/entries.

## Validation

- Live capture of the charger's handshakes in all three of its modes (1.6J / 2.0.1 /
  auto) via a standalone CSMS, which is how the transient 1.6 probe and the per-mode
  subprotocol offers were identified.
- Deployed to a live Home Assistant 2026.7.4 with the physical charger: switching the
  charger between 1.6J, 2.0.1 and auto — including six connections inside one minute —
  logged a rebuild on every version change and completed post-connect setup on the
  first attempt each time, with no `FormatViolationError` and no reconnect loop.
- New unit tests: deterministic server-order selection (both dual-stack client offer
  orders select 1.6), pin overriding that preference, pin restriction +
  no-subprotocol rejection, `on_connect` rebuild-on-version-change and
  reuse-when-unchanged, reconfigure flow round-trip.
- Full test suite passes (coverage stays ≥95%); ruff clean.
- Independently code-reviewed across several adversarial passes; the rebuild and the
  no-subprotocol rejection were additionally verified through real loopback websocket
  connections.

## Note

A companion fix for the OCPP 2.0.1 `units_changed` / connector-count symptom also
described in #2008 is being prepared as a separate PR against `ocppv201.py`; the two
changes touch disjoint files and are independent.

## Summary by CodeRabbit

* **New Features**
  * Added configurable OCPP version selection, including automatic negotiation and pinned OCPP 1.6 or 2.0 modes.
  * Improved connection handling when chargers reconnect using a different OCPP version.
  * Added support for reconfiguring existing OCPP connections without recreating them.

* **Bug Fixes**
  * Improved subprotocol negotiation and connection recovery for unsupported or changed versions.

* **Tests**
  * Added coverage for version negotiation, reconnection behavior, and configuration reconfiguration.


KingHavok:
The CodeRabbit finding about suppressing exceptions around `stop()` is fixed in
`7223ea1`: the rebuild path now cancels the stale ChargePoint's tasks itself if
`stop()` raises, with a regression test covering it
(`test_on_connect_cancels_tasks_when_stop_fails`). `self.tasks` is `None` until
`run()` populates it, so the cancellation guards against that too.

A note on the root cause, since it isn't specific to this PR:

`ChargePoint.stop()` closes the websocket *before* cancelling its tasks:

```python
# custom_components/ocpp/chargepoint.py:560
async def stop(self):
    """Close connection and cancel ongoing tasks."""
    self.status = STATE_UNAVAILABLE
    if self._connection.state is State.OPEN:
        await self._connection.close()   # if this raises...
    for task in self.tasks:              # ...this never runs
        task.cancel()
```

So any caller whose `close()` fails leaves the instance's `monitor_connection`
task running. That applies to the two pre-existing call sites as well — the
`finally` in `run()` (line 558) and `reconnect()` (line 573) — not only the
rebuild path added here.

I've handled it defensively at the new call site in `api.py` instead of
changing `stop()`, to keep this PR limited to negotiation/config files rather
than widening it into `chargepoint.py`.

Happy to do either of the following, whichever you prefer:

- move the fix into `stop()` in this PR (a `try/finally` around the close, a
  few lines), so all call sites benefit; or
- leave `stop()` alone here and open a separate PR for it.

Just say which and I'll follow up.

KingHavok:
The CodeRabbit finding about suppressing exceptions around `stop()` is fixed in
`7223ea1`: the rebuild path now cancels the stale ChargePoint's tasks itself if
`stop()` raises, with a regression test covering it
(`test_on_connect_cancels_tasks_when_stop_fails`). `self.tasks` is `None` until
`run()` populates it, so the cancellation guards against that too.

A note on the root cause, since it isn't specific to this PR:

`ChargePoint.stop()` closes the websocket *before* cancelling its tasks:

```python
# custom_components/ocpp/chargepoint.py:560
async def stop(self):
    """Close connection and cancel ongoing tasks."""
    self.status = STATE_UNAVAILABLE
    if self._connection.state is State.OPEN:
        await self._connection.close()   # if this raises...
    for task in self.tasks:              # ...this never runs
        task.cancel()
```

So any caller whose `close()` fails leaves the instance's `monitor_connection`
task running. That applies to the two pre-existing call sites as well — the
`finally` in `run()` (line 558) and `reconnect()` (line 573) — not only the
rebuild path added here.

I've handled it defensively at the new call site in `api.py` instead of
changing `stop()`, to keep this PR limited to negotiation/config files rather
than widening it into `chargepoint.py`.

Happy to do either of the following, whichever you prefer:

- move the fix into `stop()` in this PR (a `try/finally` around the close, a
  few lines), so all call sites benefit; or
- leave `stop()` alone here and open a separate PR for it.

Just say which and I'll follow up.

drc38:
> The CodeRabbit finding about suppressing exceptions around `stop()` is fixed in `7223ea1`: the rebuild path now cancels the stale ChargePoint's tasks itself if `stop()` raises, with a regression test covering it (`test_on_connect_cancels_tasks_when_stop_fails`). `self.tasks` is `None` until `run()` populates it, so the cancellation guards against that too.
> 
> A note on the root cause, since it isn't specific to this PR:
> 
> `ChargePoint.stop()` closes the websocket _before_ cancelling its tasks:
> 
> ```python
> # custom_components/ocpp/chargepoint.py:560
> async def stop(self):
>     """Close connection and cancel ongoing tasks."""
>     self.status = STATE_UNAVAILABLE
>     if self._connection.state is State.OPEN:
>         await self._connection.close()   # if this raises...
>     for task in self.tasks:              # ...this never runs
>         task.cancel()
> ```
> 
> So any caller whose `close()` fails leaves the instance's `monitor_connection` task running. That applies to the two pre-existing call sites as well — the `finally` in `run()` (line 558) and `reconnect()` (line 573) — not only the rebuild path added here.
> 
> I've handled it defensively at the new call site in `api.py` instead of changing `stop()`, to keep this PR limited to negotiation/config files rather than widening it into `chargepoint.py`.
> 
> Happy to do either of the following, whichever you prefer:
> 
> * move the fix into `stop()` in this PR (a `try/finally` around the close, a
>   few lines), so all call sites benefit; or
> * leave `stop()` alone here and open a separate PR for it.
> 
> Just say which and I'll follow up.

Thanks, a separate PR is preferred for the stop() fix

drc38 (approved):

PR #2020, opened 2026-07-30
Handle station-level StatusNotification in OCPP 2.0.1

KingHavok:
Part of the OCPP 2.0.1 work from #2008 (follows #2012 and #2013). One more PR — a connector-count fallback for chargers whose inventory reports no connectors — builds on this and will follow separately.

## The problem

Some chargers send a **station-level** `StatusNotification` with `evseId=0` / `connectorId=0` — the FoxESS A-series sends one on every boot, mirroring OCPP 1.6's `connectorId=0` convention for the charge point itself.

`_apply_status_notification` feeds it straight into the per-connector bookkeeping, where `evse_id - 1 == -1`:

- on the first such notification, while `_connector_status` is still empty, `self._connector_status[-1]` raises `IndexError`, which propagates out of the status-notification flush and **aborts `post_connect`** — the charger never finishes setting up;
- once `_connector_status` is populated, Python's negative indexing makes it **silently write the station's status into the LAST EVSE's slot**.

## The fix

Station-level notifications are recorded as the charger-level `Status` metric — the same key the OCPP 1.6 handler uses for `connectorId=0` and that the availability switch reads — and never enter the per-connector path.

Deliberately **not** written to `(0, Status.Connector)`: that key is owned by the EVSE aggregation in `_report_evse_status`, and a station-level `Available` written there would mask a faulted connector through the flattened sensor's fallback chain.

## Testing

- Regression test added to the main v201 flow: a station-level `Unavailable` arrives before the connector's `Available`, and the test asserts no crash, the charger-level `Status` metric receives it, and the connector's own status is not clobbered. Mutation-verified: with the guard removed, the flush aborts and connector status never lands.
- Validated on live hardware (FoxESS A022KP1, HA 2026.7.4): the charger's boot-time `(0,0)` notification routes to the charger `Status` sensor on every 2.0.1 connection, with normal connector handling unchanged.
- Full suite and `pre-commit run --all-files` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OCPP 2.0.1/2.1 station-level `StatusNotification` handling to apply correctly for the valid `(0,0)` case.
  * Prevented misattribution of status when notifications contain malformed or degenerate EVSE/connector identifiers; such updates are now dropped safely.
  * Ensured station-level availability is tracked independently without overwriting EVSE-level aggregated connector status.
* **Tests**
  * Added unit coverage for station-level status behavior, including immediate handling without a connector map and validation that malformed identifiers don’t update metrics.

### DIFF
--- a/custom_components/ocpp/ocppv201.py
+++ b/custom_components/ocpp/ocppv201.py
@@ -165,6 +165,24 @@ class ChargePoint(cp):
         self, timestamp: str, connector_status: str, evse_id: int, connector_id: int
     ):
         """Update per connector and evse aggregated."""
+        # Station-level notifications (evseId=0 / connectorId=0, which the OCPP
+        # 2.0.1 spec allows and e.g. the FoxESS A-series sends on every boot)
+        # don't belong in the per-connector bookkeeping below: evse_id - 1 == -1
+        # would either raise IndexError, on the first such notification when
+        # _connector_status is still empty, or silently write the station's
+        # status into the LAST EVSE's slot once it is not.
+        # Record them as the charger-level Status metric instead - the same key
+        # the OCPP 1.6 handler uses for connectorId=0 and that the availability
+        # switch reads. (0, Status.Connector) must stay owned by the EVSE
+        # aggregation in _report_evse_status, or a station-level 'Available'
+        # would mask a faulted connector via the flattened sensor's fallback
+        # chain.
+        if evse_id < 1 or connector_id < 1:
+            self._metrics[(0, cstat.status.value)].value = ConnectorStatusEnumType(
+                connector_status
+            ).value
+            return
+
         if evse_id > len(self._connector_status):
             needed = evse_id - len(self._connector_status)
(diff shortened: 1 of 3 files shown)


KingHavok:
Both findings addressed in c5d0b97. On the predicate: took the intent but not the committable suggestion — restricting to `== (0, 0)` alone would route (1, 0)/(0, 1) into the per-connector path, where `connector_id - 1 == -1` indexes negatively: the same IndexError/last-slot-overwrite class this PR fixes (applying the suggestion verbatim fails the new parametrized test with IndexError). Instead: exact (0, 0) → charger-level Status; other degenerate ids are debug-logged and dropped (neither station-level nor a real connector — recording them as station status would misattribute); real pairs unchanged. Sub-1 ids are also exempt from map buffering per the first comment, with regression tests covering the no-map path.

KingHavok:
@coderabbitai review

drc38 (approved):

PR #2022, opened 2026-07-30
Feature: Add permanent cable lock control

RobWaterhouse95:
Background
I have the Rolec EVO EV Charger, it has functionality on the Rolec app to toggle whether the charging cable if permanently locked or not (Permanent Cable Lock). This configuration is surfaced to OCPP via the UnlockConnectorOnEVSideDisconnect key.

Solution
Add a switch to the Controls section of the OPCC charger integration.

Notes
AI was used to create this change (I know .NET, not Python)
I have tested the change on my live Home Assistant integration, changing the values from the app reflects in Home Assistant and vice versa.

## Summary by CodeRabbit

* **New Features**
  * Added a **Permanent Cable Lock** control for supported Rolec EVO chargers.
  * The control lets you enable or disable permanent cable locking through the charger’s OCPP configuration.
  * Unsupported chargers no longer display this control.

* **Bug Fixes**
  * Improved handling of unavailable or unknown charger configuration values.
  * Configuration-based controls now accurately reflect their current state.

### DIFF
--- a/custom_components/ocpp/switch.py
+++ b/custom_components/ocpp/switch.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+import logging
 from typing import Final
 
 from homeassistant.components.switch import (
@@ -17,6 +18,8 @@ from homeassistant.helpers.entity import DeviceInfo
 from homeassistant.util import slugify
 from ocpp.v16.enums import ChargePointStatus
 
+_LOGGER = logging.getLogger(__package__)
+
 from .api import CentralSystem
 from .const import (
     CONF_CPID,
@@ -27,7 +30,7 @@ from .const import (
     DOMAIN,
     ICON,
 )
-from .enums import HAChargerServices, HAChargerStatuses
+from .enums import ConfigurationKey, HAChargerServices, HAChargerStatuses
 
 
 # Switch configuration definitions
@@ -43,6 +46,9 @@ class OcppSwitchDescription(SwitchEntityDescription):
     metric_condition: list[str] | None = None
     default_state: bool = False
     per_connector: bool = False
+    config_key: str | None = None
+    config_on_value: str | None = None
+    config_off_value: str | None = None
 
 
 SWITCHES: Final[list[OcppSwitchDescription]] = [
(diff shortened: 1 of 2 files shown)


drc38:
Hi @RobWaterhouse95 as this is a specific charger feature and not universal to ocpp it won't be incorporated into this repo.

You can achieve similar functionality using HAs template and python script features.

RobWaterhouse95:
UnlockConnectorOnEVSideDisconnect isn't charger specific, it is part of the OCPP standard?

drc38:
> UnlockConnectorOnEVSideDisconnect isn't charger specific, it is part of the OCPP standard?

It only applies to socketed chargers and is irrelevant for tethered.

RobWaterhouse95:
In my view, given that the switch is only created when the charger advertises support, this feels like a fairly safe change. It should not introduce a meaningful risk of misconfiguration unless a vendor is incorrectly surfacing the configuration key.

> > UnlockConnectorOnEVSideDisconnect isn't charger specific, it is part of the OCPP standard?
> 
> It only applies to socketed chargers and is irrelevant for tethered.

The setting in question is part of the OCPP standard, and while it is only relevant for socketed chargers, that still makes it a valid integration target. It seems sensible for the integration to support it once, rather than requiring each user with a compatible charger to implement the same behaviour manually.

drc38:
What is your use case for needing to alter its value continually with a switch, rather than set and forget?

RobWaterhouse95:
I have noticed my charger cable is sometimes not locked, for unknown reasons. It would be nice to have the ability to lock it in the affirmative to ensure my cable cannot be stolen. My only guess is that it happens after the charging session finishes. I have yet to confirm that.

I also wanted to give back to the project in whatever way I could, I was hoping this would be a good first step, instead of asking someone else to fix it for me.

If you still think this is best placed as a custom automation then so be it. I just wanted to help out and implement, in my view, OCPP features. I understand if you don't think it fits within scope.

drc38:
Thanks for the extra detail, take a look at your ocpp configuration variables there should be 1 you can set to keep the cable locked after charging finishes.

PR #2024, opened 2026-07-30
Give the inventory attempt a single owner

KingHavok:
Follow-up to #2023: this was the fix for CodeRabbit's second-round finding there, but the PR merged while it was being pushed, so it never entered the merge.

## The problem

Before the first `NotifyReport` part creates `_inventory`, two concurrent `_get_inventory()` calls — the boot notification's `post_connect` racing the 10s monitor backstop's — both pass the cached-inventory guard. The second overwrites the owner's `_wait_inventory` event, and once either attempt settles and clears it, the other's accepted `GetBaseReport` response dereferences `None` at `self._wait_inventory.wait()` — an `AttributeError` that aborts its `post_connect`. (The race predates #2023; this closes it.)

## The fix

A second concurrent caller returns without touching the attempt: no second `GetBaseReport`, no event overwrite, and the buffered-status drain stays with the single owner's settle point. The concurrent caller proceeding early is the already-documented double-`post_connect` behaviour — now without a crash path.

Also strengthens the partial-topology regression from #2023 to report two connectors, so its count assertion distinguishes retained partial topology (2) from the one-connector floor (1).

## Testing

New regression: two concurrent `_get_inventory()` calls with a slow accepted request — asserts exactly one `GetBaseReport` goes out and the attempt settles cleanly. Mutation-verified: with the guard removed, the test fails with a second request sent. Full suite and `pre-commit run --all-files` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of simultaneous inventory requests to prevent conflicting updates.
  * Ensured connector information and buffered status notifications are retained correctly during partial inventory responses.
  * Prevented duplicate inventory requests and ensured waiting states are cleared after completion, including when an inventory attempt fails or is cancelled.
  * Ensured subsequent inventory requests can proceed after an unsuccessful attempt.

### DIFF
--- a/custom_components/ocpp/ocppv201.py
+++ b/custom_components/ocpp/ocppv201.py
@@ -275,6 +275,15 @@ class ChargePoint(cp):
     async def _get_inventory(self):
         if self._inventory is not None:
             return
+        if self._wait_inventory is not None:
+            # An attempt is already in flight (post_connect can run twice:
+            # boot notification racing the 10s monitor backstop). Taking
+            # ownership here would overwrite the owner's event, and once the
+            # owner settled and cleared it, this caller's accepted response
+            # would dereference None at _wait_inventory.wait(). Return and
+            # leave the attempt - and the drain at its settle point - to the
+            # single owner.
+            return
         self._wait_inventory = asyncio.Event()
         req = call.GetBaseReport(1, "FullInventory")
         resp: call_result.GetBaseReport | None = None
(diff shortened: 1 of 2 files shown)


KingHavok:
@coderabbitai review

drc38:
@KingHavok let me know when ready to merge

KingHavok:
Ready to merge, thanks. Both CodeRabbit findings are addressed in bf254f1 — the ownership release moved into a `finally`, and the buffered-status drain moved inside it so an escaping `TimeoutError` or cancellation can't skip it. Full suite and pre-commit pass.

drc38 (approved):

PR #2025, opened 2026-07-31
Clamp a stored connector count of zero to one

KingHavok:
Follow-up to #2023, found while running that fix on real hardware.

## The problem

`post_connect` persists whatever `get_number_of_connectors()` computed into the config entry:

```python
self.num_connectors = await self.get_number_of_connectors()
...
s[CONF_NUM_CONNECTORS] = int(self.num_connectors)
```

Before #2023, an OCPP 2.0.1 charger whose `GetBaseReport` reported no connectors (the FoxESS A-series shape from #2008) made that `0` — and `0` was written to config. The per-connector entity loops in `switch`, `number`, `button` and `sensor` then run `range(1, 0 + 1)`, which is empty, so **the charger comes up with no Charge Control switch at all**. The integration cannot start a charge, and there is nothing in the UI or the log to say why: the entity is simply absent, showing only as a `restored: true` stub from the registry.

Because it is config rather than runtime state, it survives restarts, OCPP version changes and integration updates. It does self-repair on the next `post_connect` that completes — but a charger that never gets that far, one erroring during setup or simply offline, stays silently unusable. The existing migration doesn't help either: it only fills the key in when it is *missing*, not when it is present and zero.

## The fix

Clamp the stored count to at least one where the entities are built:

```diff
-                    num_connectors = int(
-                        cfg.get(CONF_NUM_CONNECTORS, DEFAULT_NUM_CONNECTORS)
-                    )
+                    num_connectors = max(
+                        1, int(cfg.get(CONF_NUM_CONNECTORS, DEFAULT_NUM_CONNECTORS))
+                    )
```

in the four platforms that build per-connector entities. #2023 stops the bad value being produced; this stops any stored value from erasing a charger's controls, and repairs an already-affected config on the next load rather than making it wait for a clean `post_connect`.

## Testing

`tests/test_zero_connector_config.py` sets up a charger whose config entry says `num_connectors: 0` and asserts the charge control switch, maximum current number and reset button all exist. Mutation-verified: with the clamp removed the test fails with `switch.<cpid>_charge_control` returning `None` — the exact real-world symptom.

Full suite and `pre-commit run --all-files` pass.

## How it was found

A FoxESS A022KP1 that had run OCPP 2.0.1 against a version predating the connector floor. Its config entry held `num_connectors: 0`; the charge control switch was missing entirely and the charger could not be started from Home Assistant. Restarting, rebooting and switching the charger between 1.6 and 2.0.1 all made no difference — the value only cleared when the config entry was corrected directly.

## Note for users recovering from a stored `0`

When the connector entities come back, session sensors regain their units — `Time.Session` goes from unit-less to `min` — so Home Assistant raises a one-time `units_changed` repair against the unit-less statistics recorded during the affected period. Choosing **"Update the unit of the historic statistic values, without converting"** relabels the history and keeps it; the values were always minutes, only the label was missing while the connector slots were uninitialised.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where devices configured with zero connectors would not create essential control and monitoring entities. Charge control buttons, current adjustments, sensor readings, and reset functions now properly appear regardless of connector count configuration.

### DIFF
--- a/custom_components/ocpp/button.py
+++ b/custom_components/ocpp/button.py
@@ -71,8 +71,14 @@ async def async_setup_entry(hass, entry, async_add_devices):
         for item in entry.data.get(CONF_CPIDS, []):
             for _, cfg in item.items():
                 if cfg.get(CONF_CPID) == cpid:
-                    num_connectors = int(
-                        cfg.get(CONF_NUM_CONNECTORS, DEFAULT_NUM_CONNECTORS)
+                    # Clamp to at least one connector: a stored count of 0
+                    # would make the entity loops below empty, so the charger
+                    # would come up with no connector entities at all and no
+                    # indication why. Config can hold 0 from an integration
+                    # version whose OCPP 2.0.1 inventory handling yielded no
+                    # connectors - post_connect persists whatever it computed.
+                    num_connectors = max(
+                        1, int(cfg.get(CONF_NUM_CONNECTORS, DEFAULT_NUM_CONNECTORS))
                     )
                     break
             else:
--- a/custom_components/ocpp/number.py
+++ b/custom_components/ocpp/number.py
@@ -73,8 +73,14 @@ async def async_setup_entry(hass, entry, async_add_devices):
         for item in entry.data.get(CONF_CPIDS, []):
             for _, cfg in item.items():
                 if cfg.get(CONF_CPID) == cpid:
-                    num_connectors = int(
-                        cfg.get(CONF_NUM_CONNECTORS, DEFAULT_NUM_CONNECTORS)
+                    # Clamp to at least one connector: a stored count of 0
+                    # would make the entity loops below empty, so the charger
+                    # would come up with no connector entities at all and no
+                    # indication why. Config can hold 0 from an integration
+                    # version whose OCPP 2.0.1 inventory handling yielded no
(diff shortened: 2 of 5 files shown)


drc38:
If removing the integration/adding it back or reloading it fixes the problem with your earlier fixes this PR does not seem necessary?

KingHavok:
You're right that it self-heals: with #2023 in place `get_number_of_connectors()` can't return 0 on 2.0.1, the 1.6 path always returns at least 1, and any completed `post_connect` rewrites the value and triggers a reload. Re-adding the integration fixes it too.

The narrow case I hit is that the platforms build their connector entities in `async_setup_entry`, which runs before `post_connect` — so a config already holding 0 produces no Charge Control entity until a charger connects *and* completes setup. Mine didn't (it was erroring during setup, then offline), so it stayed unusable for hours with nothing in the log pointing at the cause; I only found it by reading `.storage`. The clamp just makes the entities independent of how the stored value got there.

That said, it's belt-and-braces over a fix you've already merged, and it's your call — happy to close it if you'd rather keep the surface small.

drc38:
Prefer to address in readthedocs than code. I see there is an option to delete devices that could be incorporated too https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/stale-devices/

PR #2027, opened 2026-08-01
docs: troubleshooting for a charger with no connector entities

KingHavok:
Follows #2025 — you preferred to address this in readthedocs rather than code, so here it is as docs.

Adds a Troubleshooting section to `docs/debugging.md` for a charger left with no connector entities because the stored connector count is `0`: no Charge Control switch, so charging can't be started or stopped from Home Assistant.

The reason it's worth documenting is that the user gets nothing to go on. The entity is simply absent (or a `restored: true` registry stub), there's no error in the log, and restarting, reloading and switching OCPP versions all appear to do nothing — because the value is in the config entry, not runtime state. I only found it by reading `.storage`.

The section covers the symptom, the cause, a one-liner to check the stored value, and the fixes — reconnecting the charger and re-adding the integration first, since #2023 means a completed `post_connect` now rewrites the value, with a direct repair of the stored value as a fallback for a charger that can't complete setup. It also notes the one-time `units_changed` repair that appears once the connector entities return, and which option keeps the existing history.

Docs only, no code changes. `debugging` is already in the `index.rst` toctree so no index change was needed.

## Summary by CodeRabbit

* **Documentation**
  * Added troubleshooting guidance for missing charger control entities when the stored connector count is zero.
  * Documented symptoms, diagnostic commands, and automatic recovery options.
  * Added reinstallation steps for cases where automatic recovery is unsuccessful.
  * Clarified the expected session-sensor unit repair after charger entities are recreated.

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -93,9 +93,6 @@ Filtering for websockets.server should yield something like this:
 2022-03-16 16:35:40 DEBUG (MainThread) [websockets.server] > TEXT '[3,"5191e2e7-f555-48b3-8b08-626679df5a80",{}]' [45 bytes]
 ```
 
-Troubleshooting
-===============
-
 No Charge Control switch, and the charger cannot be started
 -----------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/debugging.md
```diff
@@ -92,3 +92,81 @@ Filtering for websockets.server should yield something like this:
...
+```
+
+A charger reporting `0` is affected.
+
+**Fix it.** Current versions correct the value automatically the next time a
+charger connects and completes its setup, so reconnecting the charger is
+usually enough. If the charger cannot complete setup, removing the
+integration and adding it back also clears it.
+
+To repair the stored value directly instead, back it up first, then edit and
```
drc38: Lines 132 to 165 are not needed as editing the config_entries file directly is not recommended

drc38 (changes requested):

KingHavok:
Removed — the manual edit and its recovery notes are gone. What's left is the diagnosis and the two supported paths: reconnect the charger (current versions correct the value once setup completes), or remove and re-add the integration.

I also dropped the redundant "Troubleshooting" heading, which had made this the only doc with two top-level headings. Happy to put it back if you'd rather keep the grouping.

drc38 (approved):

drc38:
@KingHavok looks like readthedocs yaml needs updating as the build fails, see https://docs.readthedocs.com/platform/stable/config-file/v2.html 

Appreciate if you can do a PR as if I do I cannot review/merge. Thanks!

Issue #1991, opened 2026-07-17
Problem with more than one charger connected to a central


henrikskafte:
Bug: Multiple Charge Points are created with identical internal cpid ("charger")
Environment
Home Assistant 2026.7.2
OCPP integration v0.10.15
Two OCPP 1.6J charge points
Each charger has a unique Charge Point ID (CPID)

Example:

RATIO_P00000000004199
RATIO_P00000000012534
Problem

When configuring multiple charge points on the same OCPP Central System, the integration stores the internal configuration with:

"cpid": "charger"

for every discovered charger.

Although the WebSocket path and BootNotification correctly identify each charger with its unique CPID, the configuration keeps using the same internal value (charger).

This causes:

devices being overwritten
entities belonging to the wrong charger
firmware information being mixed
duplicate entity problems
only one functional charger appearing in Home Assistant
BootNotification is correct

The log clearly shows that both chargers connect correctly.

Example:

Websocket path:
/RATIO_P00000000004199

BootNotification accepted

charge_point_serial_number:
P00000000004199

and

Websocket path:
/RATIO_P00000000012534

BootNotification accepted

charge_point_serial_number:
P00000000012534

Therefore the problem is not caused by duplicate Charge Point IDs.

Root cause

The generated core.config_entries contains:

"RATIO_P00000000004199": {
    "cpid": "charger"
}

"RATIO_P00000000012534": {
    "cpid": "charger"
}

Both entries use the same internal identifier.

Changing them manually to

"RATIO_P00000000004199": {
    "cpid": "RATIO_P00000000004199"
}

"RATIO_P00000000012534": {
    "cpid": "RATIO_P00000000012534"
}

followed by recreating the Central System completely solves the issue.

After this modification:

both chargers appear correctly
each charger has its own entities
firmware versions remain separated
live measurements update independently
no device information is mixed
Expected behaviour

The integration should automatically use the discovered Charge Point ID as the internal cpid.

Instead of

"cpid": "charger"

it should generate

"cpid": "<actual Charge Point ID>"

for every discovered charger.

Additional note

This is not Ratio-specific.

The problem should affect any installation with more than one OCPP charge point if automatic discovery assigns the same internal cpid ("charger") to every charger.

Issue #1884, opened 2026-02-15
Entities become corrupted and stop updating when adding multiple chargers


Cybgak:
**Describe the bug**
When adding a second charger the entities of the first charger are changed and cease functioning correctly

**To Reproduce**
Completely stock install of HAOS 17.1 with HA 2026.2.2
Install HACS and then OCPP 10.10
Add first charger (EVCP-7KW-S-1PH) , measurands will be "normal" current_offered, current_import, voltage etc.
Add second identical charger measurands for second charger will be "normal" measurands for first charger become; current_offered_l1, current_import_l1, Voltage_L1_N etc.
Old measurands still exist as "not provided" by integration, if old entities are removed and new ones renamed to old names they do not update when Meter Values are sent from the charger despite the measurands showing as arriving in the logs.

**Expected behavior**
Correctly named measurands that update when data arrives from the charge point.

**Screenshots**
Initial setup with one charger:

<img width="1036" height="520" alt="Image" src="https://github.com/user-attachments/assets/9cfd3414-0fd0-4260-92af-bb7d2e65d085" />

36 entities all correctly named:

<img width="1223" height="1834" alt="Image" src="https://github.com/user-attachments/assets/c3e59ecf-3488-4cea-b6b8-b9274b7d253c" />

Add second charger and "slow1" now has 42 entities unlike slow2 which has the correct 36:

<img width="1050" height="485" alt="Image" src="https://github.com/user-attachments/assets/9b2fd3af-0a49-48e1-a513-1417bbac0e6c" />

First charger entities now changed:

<img width="1011" height="1937" alt="Image" src="https://github.com/user-attachments/assets/92a997fe-d580-49a3-862e-109a12f3c566" />

**Desktop (please complete the following information):**
Not applicable

**Smartphone (please complete the following information):**
Not applicable

**Additional context**
It appears the attributes of the initial correct measurands/entities become entities for some reason.
I have tried adding more of the same chargers and each time the previous ones become corrupt but the one added last stays useable.

I've also attached the logs with debug enabled during the adding of the second charger:

[home-assistant_2026-02-15T16-13-58.295Z.log](https://github.com/user-attachments/files/25326517/home-assistant_2026-02-15T16-13-58.295Z.log)

jthun:
Are your chargers uniquely named in their own config UIs?

Cybgak:
Tested on 0.10.13 and still the same behaviour

ric866:
[...]

Issue #1658, opened 2025-07-15
Platform ocpp does not generate unique IDs.


StefanoGiu:
**Describe the bug**
ERRORS in the log

**To Reproduce**
Errors at the start of HA

**Expected behavior**
Manage the unique id

**Additional context**
2025-07-15 09:32:42.131 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.power.reactive.export.sensor already exists - ignoring sensor.charger_power_reactive_export
2025-07-15 09:32:42.131 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.power.active.export.sensor already exists - ignoring sensor.charger_power_active_export
2025-07-15 09:32:42.131 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.energy.reactive.export.register.sensor already exists - ignoring sensor.charger_energy_reactive_export_register
2025-07-15 09:32:42.132 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.energy.active.import.register.sensor already exists - ignoring sensor.charger_energy_active_import_register
2025-07-15 09:32:42.132 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.voltage.sensor already exists - ignoring sensor.charger_voltage
2025-07-15 09:32:42.132 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.energy.session.sensor already exists - ignoring sensor.charger_energy_session
2025-07-15 09:32:42.132 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.power.active.import.sensor already exists - ignoring sensor.charger_power_active_import
2025-07-15 09:32:42.132 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.soc.sensor already exists - ignoring sensor.charger_soc
2025-07-15 09:32:42.132 ERROR (MainThread) [homeassistant.components.sensor] Platform ocpp does not generate unique IDs. ID ocpp.charger.energy.active.export.register.sensor already exists - ignoring sensor.charger_energy_active_export_register
[...]

PR #2037, opened 2026-08-05
Fix duplicate cpid validation in charge point config flow

proscar87:
## Symptom

After adding a second charge point, HA logs a burst of ~36 errors like:

```
Platform ocpp does not generate unique IDs. ID ocpp.charger.power.active.import.sensor already exists - ignoring sensor.charger_power_active_import
```

repeated across sensor/switch/number/button, and the second charger's entities are silently dropped or corrupt the first charger's entities. Reported in #1991 and #1884; #1658 hit the same symptom and was closed by the stale bot without a fix.

## Root cause

`custom_components/ocpp/config_flow.py`, `async_step_cp_user` (around line 207 pre-fix):

```python
if user_input is not None:
    # Don't allow duplicate cpids to be used
    self._async_abort_entries_match({CONF_CPID: user_input[CONF_CPID]})
```

`_async_abort_entries_match` compares against each config entry's **top-level** `data`. But `cpid` never lives at that level — it's nested inside `entry.data[CONF_CPIDS] = [{cp_id: {"cpid": "...", ...}}, ...]` (see `__init__.py` around line 118-127, and where `cp_user` itself builds this structure a few lines below the check, around line 217-223).

So the match dict `{CONF_CPID: "some_cpid"}` is checked against a key that never exists at the top level of `entry.data` — the check can never find a hit. **It's a no-op.** Two different charge points (different `cp_id`, the OCPP-level websocket identity) can be configured with the same `cpid` (the user-chosen friendly name), with nothing to stop it.

That matters because `cpid` — not `cp_id` — is what entity `unique_id`s are built from, integration-wide, with no per-config-entry scoping:

```python
parts = [DOMAIN, self.cpid, self.entity_description.key, SENSOR_DOMAIN]
```

(`sensor.py`; same pattern in `switch.py`, `number.py`, `button.py`). Duplicate `cpid` → duplicate `unique_id` → HA drops the second charger's entities.

`#1886` (merged) added *format* validation for cpid (regex for lowercase/digits/`_`), not *uniqueness*. That gap is still open — confirmed no other open PR addresses it as of writing.

## Fix

Replaced the no-op `_async_abort_entries_match` call with a real uniqueness check, `_other_cpids_in_use()`, that:
- walks every OCPP config entry's `CONF_CPIDS` (not just the current central system's — `unique_id` isn't scoped per entry, so two charge points on two different central systems can't share a cpid either),
- excludes the charge point currently being (re)configured (`self._cp_id`), so resubmitting its own cpid isn't flagged as a duplicate of itself,
- and on a hit, shows a form error `duplicate_cpid` instead of silently proceeding — following the same pattern `#1886` used for `invalid_cpid` (added the string to `translations/en.json` and `translations/i-default.json`, English only).

## Why this is low risk for existing installations

This only changes the **new charge point** step of the config flow (`async_step_cp_user`), which runs when a not-yet-configured `cp_id` connects (see `api.py::on_connect`, which only triggers discovery when `cp_id not in self.settings.cpids`). It does **not**:
- change how `unique_id` is computed for any existing entity,
- migrate or touch any stored config entry data,
- affect chargers that are already configured.

It's strictly tightening input validation on setup of a brand-new charge point — not the "add unique_id to a platform that didn't have one" case that would actually duplicate entities in production.

## What I validated

- Confirmed via `gh pr list`/`gh issue view` that no open PR currently covers #1991 or #1884, and #1658 is closed unresolved by the stale bot.
- Read `__init__.py`, `api.py`, `sensor.py`/`switch.py`/`number.py`/`button.py` to confirm the `unique_id` construction and the `on_connect` discovery-trigger condition described above.
- Added tests to `tests/test_config_flow.py`:
  - `test_duplicate_cpid_value_rejected`: two different charge points (different `cp_id`) reusing the same `cpid` is rejected with `errors["base"] == "duplicate_cpid"`; a genuinely unique `cpid` on retry succeeds.
  - `test_duplicate_cpid_value_rejected_across_entries`: the same rejection holds across two different central-system config entries.
  - `test_reconfigure_own_cpid_not_flagged_duplicate`: resubmitting a charge point's own cpid for its own `cp_id` is not flagged as a duplicate of itself.
  - Fixed `test_successful_discovery_flow`, which was unknowingly relying on the bug — its "different CP IDs are allowed" second charge point reused the *same* `cpid` as the first and asserted success. Gave it a distinct cpid, which is the behavior that should actually be required.
- Ran the full suite (`pytest tests`, 172 tests) before and after: all passing both times, confirming the fix doesn't regress anything else.
- Ran `pre-commit run --all-files` (pyupgrade, ruff, ruff-format): passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate charge-point identities across configuration entries.
  * Rechecked identity uniqueness before saving configuration to prevent setup conflicts.
  * Allowed an existing charge point to resubmit its own identity without triggering a conflict.
  * Preserved existing identity format validation.
  * Added localized messages for invalid and duplicate identities.

* **Tests**
  * Expanded coverage for duplicate validation within and across entries, including setup-time race conditions.

### DIFF
--- a/custom_components/ocpp/config_flow.py
+++ b/custom_components/ocpp/config_flow.py
@@ -196,6 +196,26 @@ class ConfigFlow(ConfigFlow, domain=DOMAIN):
         self._abort_if_unique_id_configured()
         return await self.async_step_cp_user()
 
+    def _other_cpids_in_use(self) -> set[str]:
+        """Return every cpid already configured for a charge point other than the current one.
+
+        cpid (as opposed to cp_id, the OCPP-level charge point identity) is
+        user-chosen and is what entities' unique_id is built from
+        (DOMAIN.cpid.key...), so it must stay unique across every charge
+        point of every OCPP config entry, not just within the current
+        central system. The charge point currently being (re)configured
+        (self._cp_id) is excluded so re-submitting its own cpid is not
+        flagged as a duplicate of itself.
+        """
+        cpids: set[str] = set()
+        for entry in self.hass.config_entries.async_entries(DOMAIN):
+            for cp_data in entry.data.get(CONF_CPIDS, []):
+                for cp_id, cp_settings in cp_data.items():
+                    if cp_id == self._cp_id:
+                        continue
+                    cpids.add(cp_settings[CONF_CPID])
+        return cpids
+
     async def async_step_cp_user(
         self, user_input: dict[str, Any] | None = None
     ) -> ConfigFlowResult:
@@ -203,8 +223,6 @@ class ConfigFlow(ConfigFlow, domain=DOMAIN):
         errors: dict[str, str] = {}
 
         if user_input is not None:
-            # Don't allow duplicate cpids to be used
-            self._async_abort_entries_match({CONF_CPID: user_input[CONF_CPID]})
             # Validate cpid format against entity id requirements (lowercase letters, digits and _)
             schema = vol.Schema(
                 {vol.Required(CONF_CPID): cv.matches_regex(r"^[\da-z_]+$")}
@@ -214,6 +232,13 @@ class ConfigFlow(ConfigFlow, domain=DOMAIN):
             except vol.Invalid:
                 errors["base"] = "invalid_cpid"
             else:
+                # cpid is used to build entity unique_ids (DOMAIN.cpid.key...)
+                # across every OCPP config entry, so it must be unique
+                # integration-wide, not just within this central system.
+                if user_input[CONF_CPID] in self._other_cpids_in_use():
+                    errors["base"] = "duplicate_cpid"
+
+            if not errors:
                 cp_data = {
                     **user_input,
                     CONF_NUM_CONNECTORS: self._detected_num_connectors,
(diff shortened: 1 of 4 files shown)


drc38:
@proscar87 do the code rabbit issues need resolving before merging?

proscar87:
@drc38 Yes — both were real. I checked each against the code rather than taking them on faith, and both are now fixed with a test that fails without the fix.

**1. Self-exclusion scoped to `cp_id` alone.** `cp_id` is the OCPP-level identity the charger announces, so it isn't unique across central systems — two of them can each have a charge point called `charger`. Skipping on `cp_id` alone also skipped the *other* system's record, so its cpid could be reused. Now matched on `entry_id` **and** `cp_id`.

My existing cross-entry test didn't catch this because it used two different `cp_id`s. The new one uses the same `cp_id` on both entries, which is exactly the case that was slipping through.

**2. The gap before persisting.** With autoconfig off, the cpid is validated one step before anything is written, so another flow can take it in between.

I did **not** add a lock or a reservation for this. The check is repeated immediately before `async_update_entry`, with no `await` between the two — on the single-threaded event loop nothing can interleave there, so the window closes without new machinery. If you'd rather have an explicit reservation anyway, say so and I'll add it, but I'd rather not introduce shared state for a window that's already closed.

The test simulates it by writing the colliding cpid into the entry while the flow sits on the measurands form.

Suite: 11 passing.

drc38 (approved):

Issue #2039, opened 2026-08-06
OCPP 2.0.1: a probe timeout in get_supported_features discards the entire feature set


KingHavok:
## Describe the bug

On OCPP 2.0.1, `get_supported_features()` runs two capability probes and catches only `OCPPError`:

```python
# ocppv201.py
try:
    await self.call(fw_req)
    features |= Profiles.FW
except OCPPError as e:
    _LOGGER.info("Firmware update not supported: %s", e)

try:
    await self.call(trigger_req)
    features |= Profiles.REM
except OCPPError as e:
    _LOGGER.info("TriggerMessage not supported: %s", e)
```

The ocpp library raises **`asyncio.TimeoutError`**, not `OCPPError`, when a charger simply never answers:

```python
# ocpp/charge_point.py
except asyncio.TimeoutError:
    raise asyncio.TimeoutError(f"Waited {self._response_timeout}s for response on ...")
```

The timeout is 10s (`chargepoint.py`, `super().__init__(id, connection, 10)`). `websockets.ConnectionClosed` escapes the same way.

## Expected behavior

A charger that ignores one optional probe should lose only that profile, not the entire feature set.

## What actually happens

The exception propagates out of `get_supported_features()`, so `fetch_supported_features()` never reaches its assignment:

```python
self._attr_supported_features = await self.get_supported_features()   # raises here
self._metrics[(0, cdet.features.value)].value = self._attr_supported_features
```

`post_connect` then catches it with a bare `except Exception` and logs at DEBUG (`"post_connect aborted non-fatally: %s"`), so features remain `NONE`, no feature metric is written, and the only trace is a single DEBUG line after a 10s stall.

Anything `SMART`-gated is then unavailable — including `number.<cpid>_maximum_current`, so charge current cannot be set.

## To Reproduce

Connect a 2.0.1 charger that accepts the websocket but does not respond to `UpdateFirmware` (or `TriggerMessage`). Setup stalls ~10s, then `sensor.<cpid>_features` is never populated.

## Suggested fix

Widen both handlers, e.g. `except (OCPPError, asyncio.TimeoutError) as e:`, so a non-answering charger loses only the probed profile. A 10s stall is arguably also worth logging above INFO, since it is a different event from a clean "not supported" rejection.

The 1.6 path is unaffected — it derives profiles from `SupportedFeatureProfiles` rather than probing.

## Environment

- ocpp integration v0.10.17; same code on `main`
- Found while working on #2036 / #2038

[...]

PR #2046, opened 2026-08-07
Probe for smart charging when the charger omits Available on OCPP 2.0.1

### LINKED ISSUE #2035: OCPP 2.0.1: SmartCharging and Reservation not detected when the charger omits the optional <Ctrlr>/Available variable
# OCPP 2.0.1: SmartCharging and Reservation not detected when the charger omits the optional `<Ctrlr>/Available` variable

## Describe the bug

On OCPP 2.0.1, feature profiles are derived from explicit `Available` variables in the charger's `NotifyReport`:

```python
# ocppv201.py L827, L832
self._inventory.smart_charging_available = bool_value   # SmartChargingCtrlr/Available
self._inventory.reservation_available   = bool_value    # ReservationCtrlr/Available
```

`Available` is optional in OCPP 2.0.1. When a charger implements a feature but does not report that variable, `smart_charging_available` stays `False` and the profile is dropped at **L420-L423**.

## To Reproduce

Same charger, same physical state (plugged in, idle), switching only the protocol:

| | OCPP 1.6 | OCPP 2.0.1 |
|---|---|---|
| `sensor.<cpid>_features` | `CORE\|FW\|SMART\|RES\|REM\|AUTH` | `CORE\|FW\|REM\|AUTH` |
| `number.<cpid>_maximum_current` | `32.0` | `unavailable` |

From the log:

```
Feature profiles returned: CORE|FW|SMART|RES|REM|AUTH     <- 1.6
Feature profiles returned: CORE|FW|REM|AUTH               <- 2.0.1
```

What the charger actually reported on 2.0.1, parsed from `NotifyReport`:

```
SmartChargingCtrlr/ProfileStackLevel   attr=[8]
SmartChargingCtrlr/RateUnit            attr=[A,W]
SmartChargingCtrlr/PeriodsPerSchedule  attr=[24]
                                       <-- no "Available" variable at all
ReservationCtrlr                       <-- component absent from the report entirely

LocalAuthListCtrlr/Available           attr=[true]        <-- present
```

The charger clearly implements smart charging — it advertises `ProfileStackLevel`, `RateUnit` and `PeriodsPerSchedule` — it simply does not report `Available`. On 1.6 the same unit lists `SmartCharging` and `Reservation` in `SupportedFeatureProfiles`.

**`LocalAuthListCtrlr/Available` is the control:** it *is* reported, and `AUTH` is detected on both protocols. So the mechanism works; the failure is specific to absent-but-optional.

## Expected behavior

A charger that advertises a controller's other variables, or lists the feature on 1.6, should not be treated as lacking the feature because an optional flag is missing.

## Impact

- `number.<cpid>_maximum_current` becomes `unavailable`, so charge current cannot be set
- smart charging services are unavailable on chargers that support them
- reservation likewise

## Design decision

Options:

1. **Treat a present component as available** unless `Available` is explicitly `false` — i.e. absence means "unknown", not "no".
2. **Probe for the capability**, consistent with how this same function already detects `FW` and `REM` by attempting `UpdateFirmware` / `TriggerMessage` and catching `OCPPError` (L427-L446). An analogous `SetChargingProfile` or `GetCompositeSchedule` probe would fit the existing pattern.
3. **Honour `force_smart_charging` on 2.0.1** as an escape hatch — see the separate issue about that option having no effect on 2.0.1. That helps operators but doesn't fix detection.

Option 2 seems most in keeping with what `get_supported_features` already does, but options 1 and 2 are not mutually exclusive. Your call — happy to implement.

## Environment

- ocpp integration **v0.10.17**; same code on `main`
- Home Assistant OS
- Charger: **FoxESS A022KP1**, firmware `1.10,1.06`
- Controlled 1.6 → 2.0.1 → 1.6 switch, debug logging on, full `NotifyReport` captured

## Caveat

Confirmed on one charger. Omitting `Available` is spec-legal but I can't say how widespread it is. The point I'd like to raise is the "absent optional field treated as a definitive negative" pattern rather than a claim about prevalence.

KingHavok:
Closes #2035, scoped to SmartCharging only per your call on the issue.

> I'd simply go with 1, the reservation feature is very rarely used

**This is stacked on #2045** (the `asyncio.TimeoutError` fix for #2039) and will show that commit until it merges — happy to rebase onto `main` the moment it does. The reason for stacking rather than branching from `main`: this PR adds a *third* capability probe, and without #2045 an unanswered one takes the whole feature set down with it. Landing them the other way round would mean knowingly shipping a probe with that hole in it.

## The change

`SmartChargingCtrlr/Available` is optional, so absence is *unknown*, not *no*. The flag becomes tri-state so the two can be told apart, and only the unknown case is resolved by probing with `GetCompositeSchedule` — read-only, where `SetChargingProfile` would mutate charger state on every connect, as agreed on the issue.

An explicit `false` is a definite answer and is left alone, with no probe sent. So is an explicit `true`, and so is `force_smart_charging` — none of them gain a round trip.

## Any answer counts, not just Accepted

`Rejected` is a legitimate reply to a schedule request the charger cannot compute — no profile installed, or no grid-connection total for `evseId 0` — and says nothing about whether smart charging is implemented. Only a `CallError`, which is how an unimplemented message comes back, denies it. This also matches how the existing `UpdateFirmware` and `TriggerMessage` probes read their results.

## Reservation deliberately untouched

`ReservationCtrlr` is absent from these reports **entirely**, rather than present-but-missing `Available`, so there is nothing to resolve. There is no read-only probe for it either — `ReserveNow` would create a reservation. A test pins RES to the flag so this doesn't get quietly generalised later.

## Testing

11 tests in `tests/test_v201_smart_charging_probe.py`; 234 pass overall.

Verified by mutation — each of these fails the tests that should catch it:

| Mutation | Tests failed |
|---|---|
| tri-state collapsed back to `bool` (absence == no) | 2 |
| assume available when absent — the option you rejected | 3 |
| require `Accepted` rather than any answer | 1 |
| probe on an explicit `false` too | 2 |
| guard dropped, probe unconditionally | 4 |

## Caveat

Not yet confirmed against hardware: I know my FoxESS omits `Available`, but I haven't yet checked that it *answers* `GetCompositeSchedule`. I can put this build on the live charger and report back before you merge, if you'd rather wait for that — the whole point is that it fixes a charger I can actually test on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved smart-charging capability detection when charger information is unavailable or explicitly reports unsupported.
  * Prevented firmware and remote-status probe timeouts from interrupting connection setup or removing unrelated capabilities.
  * Improved handling and logging of capability probe timeouts, refusals, and closed connections.

* **Tests**
  * Added comprehensive coverage for smart-charging detection and capability probe timeout scenarios.

### DIFF
--- a/custom_components/ocpp/ocppv201.py
+++ b/custom_components/ocpp/ocppv201.py
@@ -70,7 +70,11 @@ class InventoryReport:
 
     evse_count: int = 0
     connector_count: list[int] = field(default_factory=list)
-    smart_charging_available: bool = False
+    # None means the charger never reported SmartChargingCtrlr/Available.
+    # The variable is optional in OCPP 2.0.1, so its absence is "unknown"
+    # rather than "no" - the two are told apart here so that only the
+    # unknown case is resolved by probing. An explicit false stands.
+    smart_charging_available: bool | None = None
     reservation_available: bool = False
     local_auth_available: bool = False
     tx_updated_measurands: list[MeasurandEnumType] = field(default_factory=list)
@@ -597,6 +601,29 @@ class ChargePoint(cp):
         if self.settings.force_smart_charging:
             _LOGGER.warning("Force Smart Charging feature profile")
             features |= Profiles.SMART
+
+        # SmartChargingCtrlr/Available is optional, so a charger can implement
+        # smart charging and never advertise it - the FoxESS A-series reports
+        # ProfileStackLevel, RateUnit and PeriodsPerSchedule but no Available.
+        # Absence is not a denial, so resolve it by asking. GetCompositeSchedule
+        # is read-only, unlike SetChargingProfile, which would mutate charger
+        # state on every connect. Any answer at all proves the message is
+        # implemented: Rejected is a legitimate reply to a schedule request the
+        # charger cannot compute, and says nothing about support. Only a
+        # CallError, which is how an unimplemented message comes back, denies it.
+        if (Profiles.SMART not in features) and (
+            self._inventory is None or self._inventory.smart_charging_available is None
+        ):
+            schedule_req = call.GetCompositeSchedule(60, 0)
+            try:
+                await self.call(schedule_req)
+                features |= Profiles.SMART
+            except OCPPError as e:
+                _LOGGER.info("Smart charging not supported: %s", e)
+            except TimeoutError:
+                _LOGGER.warning(
+                    "No response to GetCompositeSchedule probe, assuming no SMART"
+                )
 
         fw_req = call.UpdateFirmware(
             1,
--- a/custom_components/ocpp/ocppv201.py
+++ b/custom_components/ocpp/ocppv201.py
@@ -606,11 +606,19 @@ class ChargePoint(cp):
                 "signature": "☺",
             },
         )
+        # A probe that goes unanswered has to cost only its own profile. The
+        # ocpp library raises asyncio.TimeoutError rather than an OCPPError
+        # when a charger never replies, so catching OCPPError alone let that
+        # escape get_supported_features, past the assignment in
+        # fetch_supported_features, into post_connect's bare handler - leaving
+        # every profile off, including SMART, and no feature metric at all.
         try:
             await self.call(fw_req)
             features |= Profiles.FW
         except OCPPError as e:
             _LOGGER.info("Firmware update not supported: %s", e)
+        except TimeoutError:
+            _LOGGER.warning("No response to UpdateFirmware probe, assuming no FW")
 
         trigger_req = call.TriggerMessage("StatusNotification")
         try:
@@ -618,6 +626,8 @@ class ChargePoint(cp):
             features |= Profiles.REM
         except OCPPError as e:
             _LOGGER.info("TriggerMessage not supported: %s", e)
+        except TimeoutError:
+            _LOGGER.warning("No response to TriggerMessage probe, assuming no REM")
 
         return features
 
--- a/tests/test_v201_probe_timeout.py
+++ b/tests/test_v201_probe_timeout.py
@@ -1,0 +1,227 @@
+"""A capability probe that goes unanswered must cost only its own profile.
+
+get_supported_features probes the charger with UpdateFirmware and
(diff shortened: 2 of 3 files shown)


drc38 (approved):

KingHavok:
Confirmed on hardware after the merge — the probe works on the charger this was written for (FoxESS A022KP1, firmware 1.10/1.06).

```
11:03:08.837  Websocket Subprotocol matched: ocpp2.0.1
11:03:21.004  send    [2,"7b2913c4-…","GetCompositeSchedule",{"duration":60,"evseId":0}]
11:03:21.445  receive [3,"7b2913c4-…",{"status":"Accepted","schedule":{"duration":60,
              "chargingRateUnit":"A","chargingSchedulePeriod":[{"startPeriod":0,
              "limit":-1,"numberPhases":-1}],"evseId":0,
              "scheduleStart":"2026-08-08T01:03:21Z"}}]
11:03:21.716  Feature profiles returned: CORE|FW|SMART|REM|AUTH
```

Answered in **441 ms**, and `evseId: 0` was accepted — that was the part I was least sure of, since a charger with no grid-connection total could reasonably have refused the station-wide form.

The premise held on the same connect: `SmartChargingCtrlr` reported `PeriodsPerSchedule`, `ProfileStackLevel` and `RateUnit`, and **no `Available`**. I ran with `force_smart_charging` disabled, so SMART could only have come from the probe. `number.<cpid>_maximum_current` went from `unavailable` to `32.0`.

One thing worth recording for anyone else testing this: **pinning the integration to 2.0.1 is not the same as setting the charger to 2.0.1.** With the charger left on `auto` it agreed the `ocpp2.0.1` subprotocol and then carried on sending 1.6 payloads:

```
FormatViolationError: Additional properties are not allowed ('errorCode', 'status' were unexpected)
  payload={'connectorId': 0, 'errorCode': 'NoError', 'status': 'Available', …}
```

Every StatusNotification was rejected and the feature set collapsed to `CORE|REM`. Only after setting 2.0.1 in the charger's own app did a real 2.0.1 session start. Not an integration bug, but an easy way to get a misleading result.

PR #2051, opened 2026-08-08
Add files via upload

alex89rm:
Fix voltage aggregation for near-zero phase readings - Autel chargers shows 0.05 when phase voltage is zero.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved charging voltage calculations by excluding invalid low-voltage readings.
  * Increased accuracy for line-to-neutral, line-to-line, and fallback phase voltage measurements.

### DIFF
--- a/custom_components/ocpp/chargepoint.py
+++ b/custom_components/ocpp/chargepoint.py
@@ -698,6 +698,11 @@ class ChargePoint(cp):
             nonzero = [v for v in values if v != 0.0]
             return (sum(nonzero) / len(nonzero)) if nonzero else 0.0
 
+        def average_valid_voltages(values: list[float]) -> float:
+            """Average voltages above the noise floor; return 0.0 if none are valid."""
+            valid = [v for v in values if abs(v) >= 1.0]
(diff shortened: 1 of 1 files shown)


drc38:
Not needed as rounding to 1V can be done in the frontend

alex89rm:
This fix is not about rounding the displayed voltage.

The issue is that some single-phase chargers, such as the Autel MaxiCharger AC, report small residual values for the unused phases instead of exactly zero:

L1: 239.73 V
L2: 0.05 V
L3: 0.05 V

The current backend logic treats every non-zero value as a valid phase and averages all three:

(239.73 + 0.05 + 0.05) / 3 = 79.94 V

As a result, Home Assistant reports approximately 80 V instead of 240 V.

Frontend rounding cannot fix this, because the value produced by the integration is already incorrect before it reaches the frontend.

The proposed change only applies a small noise floor when determining which voltage phases are valid. With the fix, the same OCPP MeterValues correctly produce 239.73 V.

I reproduced the issue with a real Autel MaxiCharger AC and verified that the proposed change fixes it.

PR #2057, opened 2026-08-09
Record heartbeats on OCPP 2.0.1

### LINKED ISSUE #2056: OCPP 2.0.1: heartbeat sensor never updates - the handler answers the charger but records nothing
## Describe the bug

On OCPP 2.0.1, `sensor.<cpid>_heartbeat` never updates. The v201 heartbeat handler answers the charger but records nothing:

```python
# ocppv201.py
@on(Action.heartbeat)
def on_heartbeat(self, **kwargs):
    """Perform OCPP callback."""
    return call_result.Heartbeat(current_time=datetime.now(tz=UTC).isoformat())
```

The 1.6 handler writes the metric and pushes the entities before replying (`ocppv16.py`):

```python
now = datetime.now(tz=UTC)
self._metrics[0][cstat.heartbeat.value].value = now
self.hass.async_create_task(self.update(self.settings.cpid))
return call_result.Heartbeat(...)
```

There is no other writer of `cstat.heartbeat` anywhere in the codebase — only the two sensor-side readers — so on 2.0.1 the sensor shows whatever an earlier session left, indefinitely.

## Observed on hardware

FoxESS A022KP1, 2026-08-08: charger switched from 1.6 to 2.0.1 mid-morning. `sensor.charger_heartbeat` froze at the last 1.6-era value (`2026-08-08T00:41:40+00:00`) and still showed it that evening, through an entire afternoon of wire-confirmed 2.0.1 heartbeats:

```
13:19:41 receive [2,"5ff96f69-…","Heartbeat",{}]
```

The stale value looks plausible — it is a real recent-ish timestamp — which is presumably why this went unnoticed.

## Why CI never caught it

`test_cms_responses_v201` sends a real `Heartbeat` over the socket on every run — and asserts only that the *reply* parses. The metric is never read, so the test passed for the life of the bug.

## Expected behavior

`sensor.<cpid>_heartbeat` tracks the charger's heartbeats on 2.0.1 as it does on 1.6.

## Fix

PR incoming: mirror the 1.6 handler (write the metric, schedule the entity push, reply with the same instant recorded), plus a one-line assertion converting the existing integration test from blind to load-bearing.

## Environment

- ocpp integration v0.10.17 / current `main`
- Found while auditing loose ends from the recent 2.0.1 work (#2033–#2046 family)

KingHavok:
Closes #2056.

The v201 heartbeat handler answered the charger but recorded nothing, so `sensor.<cpid>_heartbeat` kept whatever an earlier session left — observed live on a FoxESS A022KP1 that switched from 1.6 to 2.0.1 mid-morning and showed the 1.6-era timestamp all day through wire-confirmed 2.0.1 heartbeats. The stale value looks plausible, which is presumably why nobody noticed.

## The fix

Mirrors the 1.6 handler in v201's own idiom: write the heartbeat metric, schedule the entity push — on an idle charger, heartbeats can be the only traffic, so without the push the sensor waits for an unrelated state change — and reply with the **same instant** that was recorded, so the charger and the sensor cannot disagree about what time we claimed. One deliberate non-mirror, now commented as such: 1.6 replies with whole seconds, 2.0.1 keeps its pre-existing `isoformat` reply (both valid RFC 3339).

## The existing test is now load-bearing

`test_cms_responses_v201` sends a real `Heartbeat` on every CI run and could not see this bug, because it parsed the reply and never read the metric. It now asserts the sensor's backing metric equals the parsed reply time — the pre-fix handler fails it. Same pattern as the reload-test conversion in #2054: assert structure, not just that the call answered.

## Testing

3 focused tests (metric moves off a stale value with the freshness bound taken from the test's own clock reads; reply and metric agree; entities pushed) plus the integration-test assertion. 253 pass overall.

Mutation-verified — each fails the tests aimed at it:

| Mutation | Killed by |
|---|---|
| metric write removed (pre-fix behaviour) | unit tests **and** `test_cms_responses_v201` |
| entity push removed | `test_the_entities_are_pushed` |
| written to a per-connector key | reply/metric agreement test |

## Noted, not done here

The entity push is full 1.6 parity: `update(cpid)` walks the device registry and force-refreshes every entity of the charger, at a rate the charger controls. That cost has run on the 1.6 path for years and is inherited unchanged — a targeted dispatch of just the heartbeat entity would serve both paths better, but it touches the shared dispatch machinery for both protocols, so it belongs in its own change if wanted rather than riding on a display fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OCPP 2.0.1 heartbeat handling so the heartbeat metric and response use the same UTC timestamp.
  * Heartbeat events now trigger timely Home Assistant entity updates.

* **Tests**
  * Added coverage verifying heartbeat timestamps, metric updates, and charger entity refreshes.

### DIFF
--- a/custom_components/ocpp/ocppv201.py
+++ b/custom_components/ocpp/ocppv201.py
@@ -958,7 +958,14 @@ class ChargePoint(cp):
     @on(Action.heartbeat)
     def on_heartbeat(self, **kwargs):
         """Perform OCPP callback."""
-        return call_result.Heartbeat(current_time=datetime.now(tz=UTC).isoformat())
+        # Mirrors the OCPP 1.6 handler: record the heartbeat and push the
+        # entities, so sensor.<cpid>_heartbeat tracks the charger. Without
+        # the write the sensor keeps whatever an earlier session left -
+        # heartbeats were answered here but recorded nowhere.
+        now = datetime.now(tz=UTC)
+        self._metrics[(0, cstat.heartbeat.value)].value = now
+        self.hass.async_create_task(self.update(self.settings.cpid))
+        return call_result.Heartbeat(current_time=now.isoformat())
 
     def _report_evse_status(
         self,
--- a/custom_components/ocpp/ocppv201.py
+++ b/custom_components/ocpp/ocppv201.py
@@ -965,6 +965,9 @@ class ChargePoint(cp):
         now = datetime.now(tz=UTC)
         self._metrics[(0, cstat.heartbeat.value)].value = now
         self.hass.async_create_task(self.update(self.settings.cpid))
+        # Deliberately not mirrored: 1.6 replies with whole seconds
+        # (strftime %H:%M:%SZ); 2.0.1 keeps its pre-existing isoformat
+        # reply, microseconds and all - both are valid RFC 3339.
         return call_result.Heartbeat(current_time=now.isoformat())
 
     def _report_evse_status(
--- a/tests/test_charge_point_v201.py
+++ b/tests/test_charge_point_v201.py
@@ -1063,10 +1063,15 @@ async def _run_test(hass: HomeAssistant, cs: CentralSystem, cp: ChargePoint):
     )
 
     heartbeat_resp: call_result.Heartbeat = await cp.call(call.Heartbeat())
-    datetime.fromisoformat(heartbeat_resp.current_time)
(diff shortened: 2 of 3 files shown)


drc38:
Please provide a separate PR for the targeted update when receiving a heartbeat

drc38 (approved):

PR #2072, opened 2026-08-12
Add Autel MaxiCharger to supported

jhenkens:
## Summary by CodeRabbit

* **Documentation**
  * Added support guidance for Autel MaxiCharger devices.
  * Documented reverse-proxy TLS certificate setup.
  * Added instructions for automating charging session start, stop, and reset actions.

### DIFF
--- a/docs/supported-devices.md
+++ b/docs/supported-devices.md
@@ -50,6 +50,32 @@ This list is based on the overview of OCPP 1.6 implementation for ABB Terra AC (
 
 ## [Alfen - Eve Single S-line](https://alfen.com/en/ev-charge-points/alfen-product-range)
 
+## [Autel MaxiCharger](https://autelenergy.us/pages/residential)
+The MaxiCharger works with the OCPP integration, but the community has run into a few quirks - see [Issue #1523](https://github.com/lbbrhzn/ocpp/issues/1523) for the full discussion.
+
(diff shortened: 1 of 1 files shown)


proscar87:
Two more data points for this section, from two units running here.

**Model as the chargers report it over BootNotification:** vendor `Autel`, model `MaxiChargerAC`, firmware `PFA0102|V0.00.00|V1.38.00||2.4.3.0`, feature profiles `CORE|FW|SMART|RES|REM|AUTH`. OCPP 1.6-J.

**Autel support told the owner these units don't support OCPP — that only the AC Smart Elite does.** That's wrong: both have been running against this integration for weeks, over plain `ws://` on the LAN, with charge control and metering working. Worth a line in a doc called *supported devices*, since a vendor saying "your model can't do this" is exactly what stops someone trying.

Related: the section as written only documents the `wss://` + reverse-proxy path, which is the harder setup and reads as though it were required. Plain `ws://` on a local network works with no certificate handling at all — the TLS instructions are worth keeping, but as the "if you're exposing it beyond the LAN" case rather than the default.

One caveat for the automation notes: `sensor.<cpid>_status` can read `Available` while a car is plugged in and simply not drawing. `sensor.<cpid>_status_connector` is the one that reflects reality — it read `SuspendedEV` in exactly that state, while the other still said `Available`. Anything gating on plugged-in state should use the connector sensor.

drc38 (approved):

PR #2094, opened 2026-08-28
Migrate (str, Enum) classes to StrEnum

alexisml:
## Summary

Follow-up to #2078, as requested: converts the seven `class X(str, Enum)`
declarations to `enum.StrEnum`, drops the `UP042` suppression that was holding
them in place, and then removes the 320 `.value` accesses that the conversion
makes redundant.

Ruff 0.16 flagged these once #2078 made the hooks actually run the pinned
version. Migrating them changes `__str__` and `__format__` for members that are
compared and serialised as strings across the integration, which was too broad
to fold into a routing fix — so it went into `lint.ignore` with a comment saying
it needed its own PR. This is that PR.

Two commits, reviewable separately: the base-class swap, then the `.value`
cleanup.

## 1. Migrate to `StrEnum`

**`custom_components/ocpp/enums.py`** — six classes now inherit `StrEnum`:
`HAChargerServices`, `HAChargerStatuses`, `HAChargerDetails`,
`HAChargerSession`, `OcppMisc`, `ConfigurationKey`. `Profiles` is an `IntFlag`
and is untouched.

**`custom_components/ocpp/chargepoint.py`** — `OcppVersion` now inherits
`StrEnum`. `SetVariableResult` is a plain `Enum` and is untouched.

**`.ruff.toml`** — the `UP042` entry and its four-line explanatory comment come
out of `lint.ignore`.

Three files, +9 / −13.

## 2. Drop the redundant `.value`

With the classes converted, `Member` and `Member.value` are interchangeable, so
the explicit accesses no longer carry meaning. All 320 are gone — 209 in
`custom_components/`, 111 in `tests/`. 19 files, +315 / −353.

This was done with an AST transform rather than a regex. For each
`X.member.value` the base is resolved through the module's import aliases
(`csvcs`, `cstat`, `cdet`, `csess`, `om`, `ckey`) and the member checked against
`__members__`, so only members of the seven converted classes are touched. Every
rewritten file is then asserted to parse to the same tree as the original with
exactly those nodes replaced, which fails loudly on a miss or an overreach.

Deliberately left alone:

- **The ocpp library's own enums.** `ChargingProfileKindType.relative.value`
  survives on lines whose neighbouring `OcppMisc` key was stripped — different
  class, and not a `StrEnum`.
- **`Metric.value`.** In `self._metrics[0][cdet.data_response].value` the inner
  enum access is stripped and the outer attribute kept.
- The 30 `.name` accesses, and `Profiles`.

## Why this is not a breaking change

A `StrEnum` member and its value are equal, hash equal, and both satisfy
`isinstance(x, str)`. `str()`, f-strings, `%s` and `+` all yield the value —
which is exactly what changes versus `(str, Enum)`, where those returned
`ClassName.member_name`.

`repr()` is the sole remaining difference. An AST scan finds **zero** `repr()`,
`!r` and `%r` sites in `custom_components/` and `tests/`. There are likewise no
`type(x) is str` and no `isinstance(x, Enum)` checks anywhere; the only
`isinstance(..., str)` uses are in `api.py`, which a `StrEnum` satisfies.

`_ConnectorAwareMetrics` dispatches on `isinstance(key, tuple)` /
`isinstance(key, int)`, so an enum key takes the same branch a plain string
does. Cross-referencing the 131 uncovered lines against enum usage turned up
exactly one — a `_metrics` lookup in `ocppv201.py` — which that dispatch covers.

The dicts in `ocppv16.py` that become wire payloads were checked end to end: a
`SetChargingProfile` built with enum keys and one built with `.value` keys
serialise to byte-identical JSON, and both pass the library's OCPP 1.6
jsonschema validation.

## Python version

`enum.StrEnum` requires Python 3.11. The integration already requires 3.11 or
newer — `datetime.UTC` is used in both `ocppv16.py` and `ocppv201.py`, and the
ruff target is `py312`. This introduces no new floor.

## Verification

```bash
pytest --cov=./ --cov-report=term --timeout=30 -n auto -p no:sugar tests
pre-commit run --all-files
```

314 passed, 0 failures. pyupgrade, `ruff check` and `ruff format` all pass, with
`UP042` now active.

Coverage is byte-identical to `main` — 3394 statements, 131 missed, 96.14%
before and after, with the same per-file breakdown. Expected: neither commit
adds or removes a statement.

## Checklist

- [x] Branched from `main`
- [x] Documentation updated — no change needed; nothing under `docs/` or in
      `README.md` references these declarations
- [x] Code lints — `pre-commit run --all-files` clean
- [x] Tested — full suite green, coverage unchanged

## Summary by CodeRabbit

* **Refactor**
  * Updated internal string-based enumerations to use the modern standard representation.
  * Preserved existing enum values and behavior for charging sessions, statuses, services, configuration, and OCPP-related settings.

* **Chores**
  * Updated code-quality checks to enforce the modern enum style.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -37,10 +37,6 @@ lint.ignore = [
     "D411",  # Missing blank line before section
     "E501",  # line too long
     "E731",  # do not assign a lambda expression, use a def
-    "UP042", # Class inherits from both str and enum.Enum. Migrating the
-             # existing (str, Enum) classes to StrEnum changes __str__ and
-             # __format__ behaviour for members that are compared and
-             # serialised as strings throughout, so it needs its own change.
 ]
 
 [lint.flake8-pytest-style]
--- a/custom_components/ocpp/__init__.py
+++ b/custom_components/ocpp/__init__.py
@@ -227,63 +227,63 @@ def _register_domain_services(hass: HomeAssistant) -> list[str]:
         await _route("handle_get_diagnostics", call)
 
     services = [
-        csvcs.service_configure.value,
-        csvcs.service_get_configuration.value,
-        csvcs.service_data_transfer.value,
-        csvcs.service_trigger_custom_message.value,
-        csvcs.service_clear_profile.value,
-        csvcs.service_set_charge_rate.value,
-        csvcs.service_update_firmware.value,
-        csvcs.service_get_diagnostics.value,
+        csvcs.service_configure,
+        csvcs.service_get_configuration,
+        csvcs.service_data_transfer,
+        csvcs.service_trigger_custom_message,
+        csvcs.service_clear_profile,
+        csvcs.service_set_charge_rate,
+        csvcs.service_update_firmware,
+        csvcs.service_get_diagnostics,
     ]
 
     hass.services.async_register(
         DOMAIN,
-        csvcs.service_configure.value,
+        csvcs.service_configure,
         _route_configure,
         CONF_SERVICE_DATA_SCHEMA,
         supports_response=SupportsResponse.OPTIONAL,
     )
     hass.services.async_register(
         DOMAIN,
-        csvcs.service_get_configuration.value,
+        csvcs.service_get_configuration,
         _route_get_configuration,
         GCONF_SERVICE_DATA_SCHEMA,
         supports_response=SupportsResponse.ONLY,
     )
     hass.services.async_register(
         DOMAIN,
-        csvcs.service_data_transfer.value,
+        csvcs.service_data_transfer,
         _route_data_transfer,
         TRANS_SERVICE_DATA_SCHEMA,
     )
     hass.services.async_register(
         DOMAIN,
-        csvcs.service_trigger_custom_message.value,
+        csvcs.service_trigger_custom_message,
         _route_trigger_custom_message,
         CUSTMSG_SERVICE_DATA_SCHEMA,
     )
     hass.services.async_register(
         DOMAIN,
-        csvcs.service_clear_profile.value,
+        csvcs.service_clear_profile,
         _route_clear_profile,
         CLEAR_PROFILE_SERVICE_DATA_SCHEMA,
     )
     hass.services.async_register(
         DOMAIN,
-        csvcs.service_set_charge_rate.value,
+        csvcs.service_set_charge_rate,
         _route_set_charge_rate,
         CHRGR_SERVICE_DATA_SCHEMA,
     )
     hass.services.async_register(
         DOMAIN,
-        csvcs.service_update_firmware.value,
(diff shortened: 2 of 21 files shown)


drc38:
Please remove the 320 .value accesses that are now redundant.

alexisml:
Pushed 80ba2e2, which removes the redundant `.value` accesses as suggested. All
320 are gone — 209 in `custom_components/`, 111 in `tests/`. The description is
updated to cover both commits.

Since CI has no type checker (pre-commit is pyupgrade + ruff, and ruff's
`select` carries no type rules), a regex sweep here felt too loose, so the
rewrite is an AST transform. For each `X.member.value` the base is resolved
through the module's import aliases (`csvcs`, `cstat`, `cdet`, `csess`, `om`,
`ckey`) and the member checked against `__members__`. Each rewritten file is
then asserted to parse to the same tree as the original with exactly those nodes
replaced, so a miss or an overreach fails loudly rather than silently.

That scoping is what keeps three nearby cases intact:

- the ocpp library's own enums — `ChargingProfileKindType.relative.value`
  survives on lines whose neighbouring `OcppMisc` key was stripped;
- `Metric.value` — in `self._metrics[0][cdet.data_response].value` the inner
  enum access goes and the outer attribute stays;
- the 30 `.name` accesses, and `Profiles`.

On safety: a `StrEnum` member and its value are equal, hash equal, and both
satisfy `isinstance(x, str)`, and `str()` / f-strings / `%s` all yield the
value. `repr()` is the only divergence, and an AST scan finds zero `repr()`,
`!r` and `%r` sites in either tree — no `type(x) is str` or
`isinstance(x, Enum)` checks either. `_ConnectorAwareMetrics` dispatches on
`isinstance(key, tuple)` / `isinstance(key, int)`, so an enum key takes the
plain-string branch.

The wire payloads in `ocppv16.py` were the part I did not want to take on
faith, since those dicts now have `StrEnum` keys. Built a `SetChargingProfile`
both ways: byte-identical JSON, and both pass the library's OCPP 1.6 jsonschema
validation.

Coverage is unchanged at 96.14% (3394 statements, 131 missed), same per-file
breakdown as `main`. I also cross-referenced those 131 uncovered lines against
enum usage — one hit, a `_metrics` lookup in `ocppv201.py`, covered by the
dispatch above.

314 passed, `pre-commit run --all-files` clean.

One caveat worth stating: `pytest-homeassistant-custom-component==0.13.355` now
requires Python >= 3.14, so I ran locally on 3.14 — same as `DEFAULT_PYTHON` in
the workflow. The baseline was measured on the same interpreter, so the
zero-delta on coverage holds regardless.

drc38 (approved):

Issue #2091, opened 2026-08-27
Docs wiki mirror is stale: WIKI_ACTION_TOKEN no longer authenticates


KingHavok:
## Symptom

The "Publish docs to Wiki" workflow has failed on **every docs push in its visible history** — at least six consecutive runs back to 2026-08-12 ([latest](https://github.com/lbbrhzn/ocpp/actions/runs/33113560490)):

```
fatal: could not read Password for 'https://***@github.com': No such device or address
##[error]Process completed with exit code 128.
```

`secrets.WIKI_ACTION_TOKEN` is resolving empty (or the PAT behind it has expired), so the authenticated URL degrades to a password prompt no runner can answer.

## Impact

The GitHub wiki mirror of `docs/` is frozen at whatever it held before the token died. Everything merged to docs since then exists on main and Read the Docs but not on the wiki, including at minimum: the FoxESS A-series supported-devices entry (#2073) and its follow-up amendments, the Autel MaxiCharger entry (#2072), and the multiple-central-systems documentation (#2078). Anyone reading the wiki is reading stale docs.

## Fix — maintainer-only

No PR can fix this: the `WIKI_ACTION_TOKEN` secret needs to be rotated (wiki pushes require a token with classic `repo` scope, or equivalent fine-grained access; the default `GITHUB_TOKEN` cannot push to the wiki repo, which is presumably why the secret exists). After rotating, re-running the most recent failed run — or any push touching `docs/` — resyncs the mirror, since the workflow rsyncs the whole tree with `--delete`.

Filing this as an issue rather than a PR precisely because only a repo admin can act on it.

drc38:
Thanks, fixed the token issue. Looks like it could still do with a bit of a tidy up though.

PR #2098, opened 2026-08-30
Tidy the docs-to-wiki publishing workflow

KingHavok:
Picking up the "could still do with a bit of a tidy up" from #2091, now that the token side is fixed.

## The mirror was carrying the Sphinx machinery

The workflow rsynced `docs/` into the wiki verbatim with only `.git` excluded, so the wiki also held `Makefile`, `conf.py`, `index.rst`, `requirements.txt` and the folder's own `README.md`. It now publishes only the Markdown pages (minus `README.md`, which describes the `docs/` folder rather than the wiki); the first run after merge purges what was mirrored before, because the sync still has whole-tree delete semantics.

## Publishing is now staged, which fixes a silent failure mode

The new page set is built and checked **before** the live mirror is touched, so a copy that fails partway leaves the wiki exactly as it was.

The copy also moved from `find -exec cp ... \;` to `xargs`, because **`find` exits 0 however many `-exec` invocations failed** (verified locally: replacing `cp` with `false` for every file still exits 0). Under the old form, one unreadable page would have been committed as a page *deletion* and the run would have reported success. A missing `Home.md` now aborts for the same reason.

## Security

The copy is `find -type f`, which is load-bearing rather than cosmetic: a symlink under `docs/` pointing at the wiki clone's `.git/config` is dereferenced by a plain `cp docs/*.md` and publishes `WIKI_ACTION_TOKEN` as a wiki page. I reproduced that before fixing it. Also added `permissions: contents: read`, `persist-credentials: false` on checkout, and a job timeout.

A `workflow_dispatch` trigger was considered and **rejected**: it would let anyone with write access run their own branch's workflow with the repository token, and a ref check inside the workflow can't help when the branch controls that file. Push-to-main stays the only trigger, as before — a manual resync is a re-run of the last run, which is how you applied the token fix.

Two things deliberately left for you, since they need repo-admin rights or are judgement calls:
- Scoping `WIKI_ACTION_TOKEN` to a `main`-restricted environment is the belt-and-braces version if `workflow_dispatch` is ever wanted back.
- The token stays in the clone URL rather than an `http.extraheader`: after the `-type f` fix it is unreachable by the publish copy, and the base64 extraheader form is **not** auto-masked in logs, which would be a worse exposure on error.

## Robustness

Clone and push retry three times with jittered backoff — a transient failure previously left the wiki stale until the next docs commit. Jitter because a GitHub-wide incident fails every repository's workflows at once and they would otherwise all retry on the same beat; the concurrency group only serialises this repo's own runs. Errors are not classified: matching git stderr is fragile, and a retried auth failure just fails three times seconds apart.

## Navigation drift the blind mirror was hiding

The wiki Home page and `docs/README.md` were both missing `multiple-central-systems`, which has been on Read the Docs since it merged. Added to both.

## Verification

The shell was extracted from this workflow file and run end to end against a real local git remote seeded with the wiki's current (untidy) state: pages synced, content updated, Sphinx files and README purged, a page deleted from `docs/` removed from the wiki, a symlink attack not published, `.git` intact, unchanged docs a clean no-op, an unreadable page aborting before the mirror is touched, the `Home.md` gate firing, and the retry loop exercised for real (two jittered sleeps, then a clean failure). The push leg against GitHub itself can only be proven by the first docs merge after this lands.

## Also worth knowing

`docs/Charge_automation.md` is published but linked from no contents page, and its three YAML examples are malformed (sibling-indented sensor fields, a broken block scalar, an unterminated quote) — a user copying them gets a config that fails validation. Deliberately not touched here: it wants a content PR that fixes and links it together, rather than this one advertising broken config. Happy to do that separately if you'd like.

Refs #2091.

## Summary by CodeRabbit

* **Documentation**
  * Added “Multiple central systems” to the documentation table of contents.
  * Improved consistency between documentation navigation pages.

* **Chores**
  * Updated documentation publishing to include only relevant Markdown pages.
  * Added safer publishing behavior, including validation, retries, and skipping unnecessary updates.

### DIFF
--- a/.github/workflows/publish_docs_to_wiki.yml
+++ b/.github/workflows/publish_docs_to_wiki.yml
@@ -6,33 +6,102 @@ on:
       - docs/**
     branches:
       - main
-      
+
+# Least privilege for the default token: this job only reads the repo.
+# The wiki push uses WIKI_ACTION_TOKEN, a separate credential.
+permissions:
+  contents: read
+
+concurrency:
+  group: wiki-publish
+  cancel-in-progress: false
+
 env:
-  USER_TOKEN: ${{ secrets.WIKI_ACTION_TOKEN }}
   USER_NAME: ocpp
   USER_EMAIL: ocpp@lbbrhzn.nl
-  OWNER: ${{ github.event.repository.owner.name }}
-  REPOSITORY_NAME: ${{ github.event.repository.name }}
 
 jobs:
   publish_docs_to_wiki:
     runs-on: ubuntu-latest
-    steps:  
+    timeout-minutes: 10
+    steps:
     - uses: actions/checkout@v7
+      with:
+        # Nothing here pushes to the main repo; do not leave the default
+        # token in the checkout's git config.
+        persist-credentials: false
 
-    - name: Pull wiki
+    - name: Stage wiki content
       run: |
-         mkdir tmp_wiki
-         cd tmp_wiki
-         git init
-         git config user.name $USER_NAME
-         git config user.email $USER_EMAIL
-         git pull https://$USER_TOKEN@github.com/$OWNER/$REPOSITORY_NAME.wiki.git
+        # Build the new page set BEFORE the live mirror is touched, so a
+        # copy that fails partway leaves the wiki as it was instead of
+        # publishing half a mirror. xargs propagates such a failure;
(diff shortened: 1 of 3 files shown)


drc38 (approved):

drc38:
Thanks a separate PR to fix docs/Charge_automation.md would be helpful

PR #2103, opened 2026-08-31
Limit NumberOfConnectors to a sane range to prevent OOM

jcellary:
## Summary

`ChargePoint.get_number_of_connectors()` (`ocppv16.py`) accepts any positive integer returned by the charger. A charger with corrupted configuration memory reported `NumberOfConnectors = 1850027108`. The integration then allocated per-connector metric structures for ~1.85 billion connectors, consuming all available RAM within ~2 minutes and getting the Home Assistant core process OOM-killed — reproducibly on every restart.

This PR clamps the value to a plausible range and falls back to the configured connector count otherwise.

## Environment

- **Charger:** Solplanet (AISWEI) Apollo 11 kW wallbox (`SOL APOLLO`, model string `ELH001111-WC5-1F2P`), firmware **3.03**, OCPP 1.6J over `ws://`
- **Home Assistant:** HAOS on Proxmox VM (6 GB RAM), core 2026.x, integration v0.10.x (current `main`)
- Single-connector charger, `num_connectors` configured as `1`

## What happened

With the default (autoconfig) measurand list, the integration sends on every connection:

```
ChangeConfiguration(MeterValuesSampledData, "Current.Export,Current.Import,Current.Offered,Energy.Active.Export.Interval,...,Voltage")   # ~330 chars
```

This charger's firmware has a buffer overflow in its configuration store: values longer than ~128 characters overwrite adjacent keys. After the write, `GetConfiguration` returned:

```
SupportedFeatureProfiles -> "nterval,Energy.Active.Import.Register,...,Power.Rea\u0001tive.Export,...,Voltage"
NumberOfConnectors       -> "1850027108"
```

Note that the value of `SupportedFeatureProfiles` is exactly the tail of the measurand CSV starting at character 128, including a stray `\u0001` byte.

`get_number_of_connectors()` parsed `1850027108`, checked only `n > 0`, and returned it. The subsequent per-connector metric initialisation then exhausted memory.

**Symptoms:** the `homeassistant` container grows from ~600 MB to >4.7 GB in ~130 s at 100% CPU, no exceptions are logged, and the kernel OOM-kills `python3`. A power-cycle of the charger clears the corruption, but the next HA restart re-triggers it because the integration re-writes the long CSV.

The vendor should fix the firmware separately, but the integration should not let a single bad value from a charger take down the whole Home Assistant instance.

## Fix

Limit the parsed number of connectors to a maximum of 10.

## Testing

- Reproduced the OOM on the affected charger with the original code (3 consecutive restarts, each OOM-killed at ~130 s).
- With this patch, the integration logs the warning, uses 1 connector, and HA runs stably; memory stays flat (~650 MB container RSS).
- After shortening the measurand list below the firmware's buffer size and power-cycling the charger, the charger reports sane values again and the patch is no longer exercised — i.e. no behavioural change for well-behaved chargers.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of charger-reported connector counts.
  - Limited supported connector counts to a maximum of 10 for improved reliability.
  - Valid positive counts are accepted and safely capped at 10 when necessary.
  - Missing or invalid values now consistently fall back to a single connector.

### DIFF
--- a/custom_components/ocpp/ocppv16.py
+++ b/custom_components/ocpp/ocppv16.py
@@ -87,7 +87,9 @@ _DEFAULT_LIMIT_WATTS = 22000
 _DEFAULT_LINE_VOLTAGE = 230.0
 _DEFAULT_PHASES = 1
 
-_MAX_CONNECTORS = 100
+# Upper bound for the connector count a charger reports: a corrupted value
+# would otherwise create a huge number of entities and hang Home Assistant.
+_MAX_CONNECTORS = 10
 
 _AMPS_UNIT_TOKENS = frozenset({"current", "a", "amp", "amps", "ampere", "amperes"})
 _WATTS_UNIT_TOKENS = frozenset({"power", "w", "watt", "watts"})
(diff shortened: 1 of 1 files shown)


drc38:
I agree with including an upper bound on the number of connectors to prevent HA crashing. Ten would suffice and it can be included with a 2 line change, returning the Max of the two values. The other charger issues such as not supporting autoconfig should be documented in readthedocs rather than additional code.

jcellary:
Makes sense @drc38 I brought it down to 4 lines, hope that's good enough 🙂

drc38:
> Makes sense @drc38 I brought it down to 4 lines, hope that's good enough 🙂

Nice one!

drc38 (approved):

PR #2107, opened 2026-09-01
ocppv16: negotiate largest accepted measurand subset when full CSV is rejected

Hard7Rock:
## Problem

Some OCPP 1.6 chargers reject the entire `MeterValuesSampledData` `ChangeConfiguration` CSV outright if it contains even one measurand they do not support, instead of accepting the subset they do support.

Confirmed on a real **Teltonika TeltoCharge EVC10100** unit: sending all 22 standard OCPP 1.6 measurands (or any list containing an unsupported one) is rejected wholesale with `Rejected`/`NotSupported`, even though the charger genuinely supports 6 of them. This matches the vendor's own documented list (wiki.teltonika-energy.com/view/OCPP_1.6_JSON_guide, 'Supported measurands'):

```
Current.Import, Current.Offered, Energy.Active.Import.Register, Power.Active.Import, Temperature, Voltage
```

Previously, `get_supported_measurands()` on rejection fell back to whatever `GetConfiguration` reported — typically just the charger's pre-existing default (`Energy.Active.Import.Register` alone) — silently losing every other measurand the charger actually supports even though the user explicitly configured them via the manual (non-autodetect) options flow.

## Fix

Adds `_negotiate_measurand_subset()`: when the full desired CSV is rejected, binary-search the token list — split in half, test each half via `ChangeConfiguration`, recurse into rejected halves, drop unsupported single tokens, and re-verify the recombined result (guards against a charger that accepts two halves individually but rejects them combined, e.g. a max-active-measurands limit). Finds the maximal accepted subset in O(log n) charger round-trips instead of requiring users to manually bisect the list by hand through the options flow (which is how this behaviour was originally diagnosed).

The autodetect path is unchanged.

## Testing

Manually verified end-to-end against the real Teltonika charger before writing this PR (options flow → charger rejects full list → previously fell back to the single default measurand). This PR adds automated coverage:

- `tests/test_measurand_negotiation.py` — 5 new cases:
  - full CSV rejected → negotiates down to the charger's real accepted subset
  - full CSV accepted → no negotiation needed (single call)
  - charger supports none of the requested measurands → empty result, no crash
  - transport error on the very first `ChangeConfiguration` attempt → negotiation still proceeds
  - cross-measurand interaction where two independently-accepted halves are rejected when recombined → falls back to one verified half rather than silently keeping an unverified, potentially-rejected combination

Full existing suite (314 pre-existing tests) still passes with no regressions; `ocppv16.py` coverage is 95% (repo threshold), `ruff check`/`ruff format --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved measurand configuration for chargers that reject unsupported combinations.
  * The integration now identifies and applies the largest supported subset instead of reverting to the charger’s default.
  * Added handling for transport errors and cases where no requested measurands are supported.

### DIFF
--- a/custom_components/ocpp/ocppv16.py
+++ b/custom_components/ocpp/ocppv16.py
@@ -279,22 +279,25 @@ class ChargePoint(cp):
                 _LOGGER.debug(
                     "'%s' measurands set manually to %s", self.id, desired_csv
                 )
-                if getattr(resp, "status", None) in cfg_ok:
-                    effective_csv = desired_csv
-                else:
-                    _LOGGER.debug(
-                        "'%s' manual measurands set not accepted (status=%s); using charger's value",
-                        self.id,
-                        getattr(resp, "status", None),
-                    )
-                    effective_csv = await self.get_configuration(key)
+                accepted_full = getattr(resp, "status", None) in cfg_ok
             except Exception as ex:
+                _LOGGER.debug("Manual measurands set raised for '%s': %s", self.id, ex)
+                accepted_full = False
+
+            if accepted_full:
+                effective_csv = desired_csv
+            else:
                 _LOGGER.debug(
-                    "Manual measurands set failed for '%s': %s; using charger's value",
+                    "'%s' manual measurands set not accepted (status=%s); "
+                    "negotiating largest accepted subset",
                     self.id,
-                    ex,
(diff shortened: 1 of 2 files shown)


drc38:
Not convinced this required, users can select manual entry to select the measurands the charger supports.

PR #2116, opened 2026-09-03
Add per-charge-point OCPP 1.6 timing controls

matiebird:
## Summary

Adds backward-compatible, per-charge-point OCPP 1.6 timing controls for:

- the heartbeat interval returned in `BootNotification.conf` and, when explicitly configured, the charger's `HeartbeatInterval` key;
- the optional charger-generated `WebSocketPingInterval` key;
- the existing `MeterValueSampleInterval`, now with read-before-write and post-write verification.

The Home Assistant central system's own WebSocket ping interval remains separate and unchanged.

## Motivation

Some chargers need different heartbeat and charger-generated WebSocket ping intervals from the historical defaults. The integration previously returned a fixed 3,600-second heartbeat interval in every OCPP 1.6 `BootNotification.conf`, read `HeartbeatInterval` without configuring it, and defined `WebSocketPingInterval` without exposing a per-charge-point option.

This change makes those charger-side timings configurable without imposing new writes on existing installations or chargers that do not support the keys.

### Heartbeat

- New charge-point configurations expose an OCPP heartbeat option with the existing `3600`-second default.
- `BootNotification.conf.interval` uses the selected per-charge-point value.
- Existing entries with no stored heartbeat option continue returning `3600`.
- A legacy entry does **not** start persistently writing `HeartbeatInterval=3600` merely because the integration was updated.
- When a heartbeat value is explicitly stored, the integration reads the charger key, writes only a mismatch, and reads it back after `Accepted` or `RebootRequired`.

### Charger-generated WebSocket ping

- Adds an optional per-charge-point `WebSocketPingInterval` setting.
- It defaults to unset.
- When unset, the integration does not read or modify that charger key.
- UI and documentation distinguish this from the central system's existing WebSocket ping option.

### Meter and idle sampling

- `MeterValueSampleInterval` continues to use the existing charging-sample option.
- It is read before writing, left alone when already correct, and read back after an accepted or reboot-required change.
- `ClockAlignedDataInterval` remains the separate charger idle/clock-aligned sampling option.

### Unsupported or unusual charger responses

Unknown, unsupported, read-only, rejected, timed-out, disconnected, or reboot-required timing outcomes are observable but non-fatal. Timing setup cannot make charging unavailable solely because a charger does not implement one of these optional configuration keys.

## Reconnect and session safety

OCPP 1.6 post-connect setup is coalesced and bound to the WebSocket session that initiated it. A stale setup task cannot send a mutation through a replacement connection after a reconnect.

The session fence covers:

- feature and connector discovery;
- heartbeat discovery;
- charger timing configuration;
- `MeterValuesSampledData` configuration;
- `ClockAlignedDataInterval` configuration;
- availability and trigger operations reached during initial setup.

Reconnect processing remains bounded to the newest session. OCPP 2.0.1 reconnect behaviour is unchanged.

## Configuration example

An affected charger can independently select, for example:

```text
OCPP heartbeat interval:              60
Charger-generated WebSocket ping:     60
Charging sample interval:             10
Charger idle sampling interval:       900
```

These are per-device choices, not protocol-wide defaults.

## Tests and quality gates

Added deterministic coverage for:

- new-entry defaults and options-flow persistence;
- legacy `3600` boot compatibility without a new persistent write;
- configured `BootNotification.conf` output;
- read-before-write, mismatch-only writes, and readback;
- `Accepted`, `RebootRequired`, read-only, rejected, unknown, timeout, and disconnect outcomes;
- optional/unset charger ping behaviour;
- repeated reconnect coalescing;
- stale-session races at the actual OCPP send boundary;
- initial discovery, measurand, idle-interval, availability, and trigger operations;
- unchanged OCPP 2.0.1 behaviour.

Verification on exact head `19cee8a6e3380f753bc06197b7b7d17d4f22b178`:

- complete split suite: **373/373 passed**;
- coverage: **96%**;
- broad focused suite: **175/175 passed**;
- Ruff, formatting, pre-commit, compile, translation JSON, secret-pattern, and `git diff --check`: passed;
- independent exact-commit review: **NO BLOCKING FINDINGS**.

The suite is split because two existing OCPP 1.6 test modules have an upstream-reproducible fixed-port interaction when run together; both parts pass independently.

## Field validation

The branch was deployed as part of a combined local build behind existing site-specific charger safeguards. On an OCPP 1.6 charger it:

- negotiated a direct Home Assistant connection;
- applied and read back `HeartbeatInterval=60`, `WebSocketPingInterval=60`, `MeterValueSampleInterval=10`, and `ClockAlignedDataInterval=900`;
- continued receiving heartbeats and pongs;
- completed harmless configuration round trips after a charger power cycle.

The timing feature worked, but the direct transport did not become stable on this charger. After the post-power-cycle direct connection was restored, Home Assistant Recorder showed the reconnect counter increasing from `5` to `9` over 1,014 seconds: four replacement connections, with intervals of approximately 173, 355, 254, and 233 seconds. The operator also observed availability changing in the UI. The available direct-path logs do not identify which side initiated each close, so this is evidence of repeated connection replacement, not proof that the charger alone initiated every disconnect.

The charger was then returned to its existing terminating WebSocket compatibility relay, after which the path remained stable apart from the intentional power-cycle close/reconnect. This PR therefore does **not** claim that timing changes alone resolve charger-specific HTTP/WebSocket transport defects. It adds generic timing control and ensures asynchronous post-connect work remains safe when reconnects occur.

## Scope

This PR contains only the generic upstream timing and connection-lifecycle work. It does not include the local charger-specific charging-profile, reset, or transaction-reconciliation customisations used by the field-test installation.


matiebird:
Closing after live field validation. The new timing controls applied and read back correctly, but the charger still produced repeated replacement connections when connected directly to Home Assistant. The existing terminating WebSocket compatibility relay remains the proven stable transport for this installation, so this proposal is being withdrawn for now.

PR #2001, opened 2026-07-24
Use timeout-triggered reconnect with command queue and replay

Mark-Hetherington:
On asyncio.TimeoutError place failed commands in queue. Newer commands replace commands of the same type. On reconnect replay queued commands

Changes:
- New CommandQueue class with FIFO ordering and coalescing logic
- New _call_with_timeout_handling() wrapper in ChargePoint to catch timeouts
- Modified reconnect() to drain queue via _replay_queue() before re-establishing connection
- Add _command_queue instance to ChargePoint.__init__

Integration:
- ocppv16.py: wrap set_charge_rate() (4 calls) and set_availability() with timeout handler
- ocppv201.py: wrap set_charge_rate() (2 calls), set_availability() (3 calls), start_transaction(), and stop_transaction() with timeout handler
- Add asyncio import to ocppv16.py for exception handling

Fixes: #1999

## Summary by CodeRabbit

- **New Features**
  - Added a per–ChargePoint FIFO command queue to preserve timed-out OCPP calls and replay them after reconnection.
  - Enhanced timeout-aware handling for charging profile updates, availability changes, and remote start/stop actions.

- **Bug Fixes**
  - Prevented missed OCPP requests by queueing failed calls on timeout, then replaying them sequentially after reconnect.
  - Improved command coalescing for charging profile updates using charging-profile purpose to avoid incorrectly dropping distinct requests.

- **Tests**
  - Added async pytest coverage for queue order, coalescing rules, and timeout-to-replay behavior.

PR #2122, opened 2026-09-06
Fix reconnect session ownership and bound retirement

matiebird:
## Problem

A retiring `run()` calls `stop()` through mutable `self._connection` and `self.tasks`. If a reconnect has already installed a replacement, the old runner's delayed `finally` can close the replacement socket, cancel its tasks, and mark the charge point unavailable.

## Changes

- Capture each session's socket and task ownership so old cleanup cannot close its replacement.
- Share teardown between stop callers and bound socket-close/child-settlement with an aggregate 10-second deadline.
- Retain and observe cancellation-resistant survivors, preventing same-object replacement until old work settles.
- Handle overlapping reconnects, repeated cancellation, child-initiated stop, and retries after retirement settles.
- Add native regression tests and document the lifecycle contract.

No command queue/replay, firmware, charging-profile or energy-policy changes are included.

## Reproduction: failing before, passing after

The decisive regression starts connection A, holds its final cleanup behind an explicit barrier, installs B, releases A's cleanup, and waits for A's runner/finalizer to finish before checking B. On the original code it fails specifically with **`stale finalizer closed replacement`**; with the fix it passes. Expected cancellation is retrieved and unexpected runner exceptions fail the test.

From this PR checkout, install the repository's test requirements in `.venv`, then run:

```sh
# Fixed candidate: expected PASS.
.venv/bin/python -m pytest tests/test_reconnect_lifecycle.py::test_stale_finalizer_cannot_close_replacement --no-cov --timeout=30 -p no:cacheprovider -o addopts='' -q

# Original upstream with explicit old cleanup-boundary adaptation.
# Production source remains unchanged; candidate tests have no fallback.
CANDIDATE="$PWD"
BASELINE=$(mktemp -d)
git archive bc5d879c6f49d90bc1ffd6d72107a4aa8dc00ead -o "$BASELINE/base.tar"
tar -xf "$BASELINE/base.tar" -C "$BASELINE"
cp tests/test_reconnect_lifecycle.py "$BASELINE/tests/test_reconnect_lifecycle.py"
python3 - "$BASELINE/tests/test_reconnect_lifecycle.py" <<'PY'
from pathlib import Path
import sys
p = Path(sys.argv[1])
s = p.read_text()
assert s.count('original = cp._stop_session') == 1
assert s.count('monkeypatch.setattr(cp, "_stop_session", paused)') == 1
s = s.replace('original = cp._stop_session', 'original = cp.stop')
s = s.replace('monkeypatch.setattr(cp, "_stop_session", paused)', 'monkeypatch.setattr(cp, "stop", paused)')
p.write_text(s)
PY
(
  cd "$BASELINE" || exit 1
  # Expected exit 1 at the stale-finalizer assertion, not a teardown failure.
  "$CANDIDATE/.venv/bin/python" -m pytest tests/test_reconnect_lifecycle.py::test_stale_finalizer_cannot_close_replacement --no-cov --timeout=30 -p no:cacheprovider -o addopts='' -q
)
```

## Verification

Results were recorded against a frozen source/test snapshot; current files were checked against that snapshot before publication:

- Original upstream with explicit old cleanup-boundary adaptation: the selected stale-finalizer regression **fails at `stale finalizer closed replacement`**. Candidate test directly targets `_stop_session`, so renames cannot silently switch the boundary.
- Fixed candidate, focused lifecycle/compatibility tests: **37 passed**.
- Full project suite: **369 passed** in 246.06 seconds; **96.20% coverage**, above the 95% gate.
- Ruff lint/format and diff whitespace checks passed.
- No pending-task destruction, unretrieved-exception or teardown-leak diagnostics were found in these runs.

```sh
.venv/bin/python -m pytest tests/test_reconnect_lifecycle.py tests/test_charge_point_core.py tests/test_more_coverage_chargepoint.py --no-cov --timeout=30 -p no:cacheprovider -o addopts='' -q
.venv/bin/python -m pytest --cov=./ --cov-report=xml --cov-report=term --timeout=30 --durations=10 -n 8 -p no:sugar -rA tests
.venv/bin/ruff check --no-cache custom_components/ocpp/chargepoint.py tests/test_reconnect_lifecycle.py tests/test_charge_point_core.py tests/test_more_coverage_chargepoint.py
.venv/bin/ruff format --check --no-cache custom_components/ocpp/chargepoint.py tests/test_reconnect_lifecycle.py tests/test_charge_point_core.py tests/test_more_coverage_chargepoint.py
```

Environment: Python 3.14.5, Home Assistant 2026.9.0b4, ocpp 2.1.0, websockets 17.1, pytest 9.0.3. The final full suite used eight workers rather than CI's automatic worker count; its OCPP 2.0.1 integration test takes about 211 seconds, so a 180-second whole-command limit is insufficient. HACS, Hassfest and pre-commit were not rerun locally for the final snapshot; hosted CI is separate from these local results.

Native tests exercise the actual integration with HA fixtures and controlled transports. The localhost WebSocket test covers real receive/Ping/Pong/Close with no application traffic; it is not physical-charger qualification. Python 3.14 cancellation behavior is covered: cleanup uses `asyncio.wait` plus explicit result retrieval rather than leaving a cancelled `shield()` wrapper to log a late inner exception.

## Review follow-up

- Addressed six CodeRabbit nits and added useful docstrings. The production-deadline test is marked `slow` but remains in default and CI runs.
- Added explicit public-API evidence that the retirement owner retrieves late task exceptions.
- Reproduced Codecov’s two uncovered changed lines: the unused fault assignment (removed) and the post-await survivor rejection. Added a regression that fails when that guard is removed.
- Local full-suite coverage intersects **80/80 changed executable `chargepoint.py` lines**, with no exclusions or threshold changes. Hosted Codecov results must rerun after publication.

## Scope and limitations for review

- Deadlines bound caller latency on a responsive event loop, not complete task/process termination. Surviving work intentionally blocks replacement on the same object.
- A rejected reconnect can consume separate budgets for old-session retirement and closing the unpublished incoming socket.
- Separately scheduled `post_connect`, external service work, inherited sends, and different-protocol object-rebuild fallback are not comprehensively fenced by this change.
- This does not claim to fix every charger disconnect or provide hardware recovery. No real charger or live HA deployment was used to qualify this exact upstream candidate.
- The separate queue/replay proposal in #2001 also touches reconnect; this PR deliberately does not add that behavior.

Prepared against upstream `main` at `bc5d879c6f49d90bc1ffd6d72107a4aa8dc00ead`.

## Summary by CodeRabbit

* **New Features**
  * Added support for setting a charge point’s station charge rate.

* **Bug Fixes**
  * Improved connection recovery and replacement handling during reconnects.
  * Prevented outdated reconnect attempts or cleanup operations from disrupting newer connections.
  * Improved shutdown reliability when connections close slowly, fail, or are cancelled.
  * Rejected unsafe replacement connections and invalidated pending reconnects when stopping.
  * Prevented delayed initial starts from restoring stale sessions after a stop or reconnect.

* **Documentation**
  * Added guidance covering reconnect lifecycle behavior, cleanup timeouts, cancellation handling, and known limitations.
  * Documented reconnect lifecycle regression testing and slow test execution.

### DIFF
--- a/custom_components/ocpp/chargepoint.py
+++ b/custom_components/ocpp/chargepoint.py
@@ -279,6 +279,8 @@ class ChargePoint(cp):
         # resolved; bounds the full-update fallback to the startup window.
         self._targeted_refresh_ready = False
         self.tasks = None
+        self._session = None
+        self._reconnect_token = None
         self._charger_reports_session_energy = False
 
         # Connector-aware, but backwards compatible:
@@ -575,8 +577,12 @@ class ChargePoint(cp):
     async def run(self, tasks):
         """Run a specified list of tasks."""
         self.tasks = [asyncio.ensure_future(task) for task in tasks]
+        # Capture ownership before yielding; a retiring run must never stop a
+        # replacement that has since overwritten self._connection/self.tasks.
+        self._session = None
+        session = self._get_session()
         try:
-            await asyncio.gather(*self.tasks)
+            await asyncio.gather(*session["tasks"])
         except TimeoutError:
             pass
         except WebSocketException as websocket_exception:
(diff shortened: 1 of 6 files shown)


custom_components/ocpp/chargepoint.py
```diff
@@ -587,32 +594,145 @@ async def run(self, tasks):
...
+                cleanup is not None
+                and cleanup.done()
+                and any(
+                    not task.done()
+                    for task in session.get("retirement", session["tasks"])
+                )
+            ):
+                self.status = STATE_UNAVAILABLE
+                raise TimeoutError(
+                    "OCPP retirement survivors still active; replacement fenced"
```
drc38, line 726: the meaning/use of "fenced" here is unclear to me, is it stating the replacement connection is blocked?

matiebird:
Addressed the pending-initial-start review finding in 450f2cd626488669b96d6cef93293b663aebd963. The reproducible yield is OCPP 1.6 transaction-store loading before base start(), rather than inside atomic run() publication. Stop and reconnect now invalidate one-shot initial-start admission before yielding, so a late initial start cannot replace session ownership or launch receive/monitor tasks.

Four regression scenarios cover stop, stop then reconnect, direct reconnect, and protocol-version rebuild through CentralSystem.on_connect(). Independent baseline sensitivity testing failed all four with "late start republished session"; candidate passes. Full local canonical suite: 519 passed, 97% coverage; Ruff lint/format checks passed. Independent scoped review: no blocking findings. Nothing merged or deployed.

drc38 (changes requested):
assuming fenced=blocked please update to use "blocked"

custom_components/ocpp/chargepoint.py
matiebird: Thanks, yes “blocked” is what I meant. I’ve updated all three timeout messages and the related comment and documentation in 480decf. The replacement connection is blocked while tasks from the old session are still active. This is a wording only change; the connection handling logic is unchanged.

drc38 (approved):

PR #2131, opened 2026-09-09
One station-wide Maximum Current per charger; remove the per-connector sliders

### LINKED ISSUE #2110: Per-connector maximum-current sliders all write one station-wide profile on multi-connector chargers
On a charger that reports more than one connector, the integration creates one `number.<cpid>_connector_N_maximum_current` slider per global connector (`number.py`, the `num_connectors > 1` branch), and each slider calls `set_charge_rate` with its own connector as `conn_id`. The managed limit those sliders send is a station-wide profile, so every slider ends up controlling the same thing:

- **OCPP 1.6:** every slider first attempts `ChargePointMaxProfile` id `1000` on connector `0`. When the charger accepts that, which is the normal case, the slider that moved last sets the limit for the whole station and the other sliders now display a value the charger is not applying. Only if the station-wide request is rejected does the path fall back to connector-scoped `TxProfile`/`TxDefaultProfile` on the requested connector, so which slider does what depends on the charger's answer.
- **OCPP 2.0.1:** the managed limit is a `ChargingStationMaxProfile`, id `1`. Today a positive `conn_id` routes it to a positive EVSE, which is the targeting bug in #2101. Once that is fixed the profile goes to EVSE 0 as the spec requires, and the sliders alias exactly as they do on 1.6: last write wins, one profile, N sliders.

So this is not introduced by #2101. It is a pre-existing mismatch between the entity model, one control per connector, and what the managed profile can express, one limit per station. #2101 only changes the 2.0.1 symptom from "rejected" to "quietly aliased".

To be precise about what the protocol can and cannot do here: OCPP does have finer control. `ChargingStationMaxProfile` and `ChargePointMaxProfile` are station-wide by definition, but Tx profiles can be scoped to an EVSE on 2.0.1 or a connector on 1.6. Only connectors that share one EVSE are inherently indistinguishable through `SetChargingProfile`. The limitation is in what the sliders currently send, not in the protocol.

**Possible directions**

- Present one station-wide slider on multi-connector chargers, since that is what the managed path actually controls, and leave per-connector control to `custom_profile`.
- Keep per-connector sliders but have them write an EVSE- or connector-scoped Tx profile, so each slider means what its name says. That takes on the session-scope and precedence questions that were deliberately kept out of #2101.
- Keep the current entities and document that they alias.

This is an entity-model decision rather than a bug with one obvious fix, so I'd rather get a steer before proposing code.

KingHavok:
Closes #2110

### Problem

On a charger with more than one connector the integration created one `Maximum Current` slider per connector, and every one of them wrote the same station-wide profile: `ChargePointMaxProfile` on connector 0 for OCPP 1.6, `ChargingStationMaxProfile` on EVSE 0 for 2.0.1. The slider moved last set the limit for the whole station while the others displayed values the charger was not applying. On 1.6 the shared path also fell back to a `TxProfile` and `TxDefaultProfile` on one connector when the station profile was rejected, so what a slider did depended on the charger's answer.

### Change

- **One `Maximum Current` per charger**, on the charger device, whatever the connector count. The multi-connector branch of the number platform is gone; the single remaining path is the one single-connector chargers already used.
- **Legacy connector sliders are removed at setup.** Registry entries are matched by config entry, integration, number domain and the exact legacy unique id for the charger being set up, so renamed and disabled entries go too, and other chargers' entries are untouched. An existing flat entry keeps its registry identity, customisations and restored value. A newly created master starts at the configured maximum without treating a removed connector slider's state as confirmed, since restore data is keyed by entity id.
- **The slider sends only a station-wide profile.** A new `set_station_charge_rate` hook sits on the base charge point: on 1.6 it sends a single `ChargePointMaxProfile` on connector 0 and reports any other status, CALLERROR or transport failure as an error, after which the slider reverts to its last confirmed value; on 2.0.1 it is the existing managed station path. There is no transaction-profile fallback in the slider, because a `TxDefaultProfile` is a per-transaction default on each connector rather than a shared ceiling, and a `TxProfile` binds one connector.
- **The `ocpp.set_charge_rate` action is unchanged**, including its 1.6 fallback chain on the connector it names.

### Breaking change

@drc38 this one wants the `breaking` label for the release notes; I can't apply labels here.

- On multi-connector chargers `number.<cpid>_connector_N_maximum_current` entities are removed and replaced by one `number.<cpid>_maximum_current`. Dashboards, scripts and automations that reference the old entities need updating.
- The slider-side fallback is removed on all 1.6 chargers, single-connector ones included. A charger that rejects `ChargePointMaxProfile` now shows an error on slider changes; the action keeps the fallback for it.
- The upgrade does not clear profiles stored on the charger. Moving the master updates only the station maximum; existing transaction or default profiles may keep imposing lower limits. `docs/Charge_automation.md` gains upgrade notes covering this, including what `ocpp.clear_profile` clears on each protocol.

### Tests

`tests/test_flat_maximum_current.py` drives the real platform, API and protocol objects: one master on the station device for one and two connectors; cleanup of stale, out-of-range, renamed and disabled legacy entries with a dotted charger id, leaving near-miss unique ids, other chargers, other platforms and other domains alone; a renamed flat entry surviving a one-to-two reconfiguration with its state; a new master ignoring restore data left by a historical flat entity or a renamed connector slider; the 1.6 slider sending exactly one station profile, including at the maximum; Rejected, NotSupported and transport failures reverting without any transaction-profile request while transactions run on both connectors; a real CALLERROR routed back through the OCPP message queue; watt-only conversion through the slider route; the action's fallback chain for an omitted, zero and explicit `conn_id`; and the 2.0.1 set, clear and refusal paths through the new hook. Existing lifecycle tests move to the flat entity id. Full suite passes locally.

## Summary by CodeRabbit

- **New Features**
  - Maximum Current is now a single station-wide control, regardless of the number of connectors.
  - The control uses the charger-level entity `number.<cpid>_maximum_current`.
  - Station-wide current limits are supported for OCPP 1.6 and OCPP 2.0.1 chargers.
  - Clearer handling is provided when a charger refuses or reports an error while applying a limit.

- **Documentation**
  - Updated guidance explains the new entity naming, station-wide behavior, profile handling, and upgrade considerations.

### DIFF
--- a/custom_components/ocpp/api.py
+++ b/custom_components/ocpp/api.py
@@ -736,13 +736,13 @@ class CentralSystem:
     async def set_max_charge_rate_amps(
         self, id: str, value: float, connector_id: int = 0
     ):
-        """Set the maximum charge rate in amps."""
+        """Set the station maximum in amps; connector_id is retained but unused."""
         # allow id to be either cpid or cp_id
         cp_id = self.cpids.get(id, id)
 
         if cp_id in self.charge_points:
-            return await self.charge_points[cp_id].set_charge_rate(
-                limit_amps=value, conn_id=connector_id
+            return await self.charge_points[cp_id].set_station_charge_rate(
+                limit_amps=value
             )
         return False
 
--- a/custom_components/ocpp/chargepoint.py
+++ b/custom_components/ocpp/chargepoint.py
@@ -422,6 +422,10 @@ class ChargePoint(cp):
         """Clear all charging profiles."""
         pass
 
+    async def set_station_charge_rate(self, limit_amps: int | float) -> bool:
+        """Set the station-wide maximum current without transaction fallbacks."""
+        raise NotImplementedError
+
     async def set_charge_rate(
         self,
         limit_amps: int | float | None = None,
--- a/custom_components/ocpp/number.py
+++ b/custom_components/ocpp/number.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 import logging
+import re
 from typing import Final
 
 from homeassistant.components.number import (
@@ -25,10 +26,8 @@ from .const import (
     CONF_CPID,
     CONF_CPIDS,
     CONF_MAX_CURRENT,
-    CONF_NUM_CONNECTORS,
     DATA_UPDATED,
     DEFAULT_MAX_CURRENT,
-    DEFAULT_NUM_CONNECTORS,
     DOMAIN,
     ICON,
 )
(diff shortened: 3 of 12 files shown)


drc38 (approved):

drc38:
@KingHavok how difficult is it to add a TxProfile slider per connector now that the id issue has been fixed?

KingHavok:
Happy to pick this up now that #2128 has cleared the way. As I said on #2110, I'd rather agree the shape before writing code: I'll flesh out the design options, from a lean version that only exists during a session and clears itself on any reconnect or restart, to a fuller one that can carry a limit across a reconnect with a liveness check first, and open a design issue with them for discussion.

PR #2147, opened 2026-09-16
ocpp16: charger level id_tag sensor not updating on StartTransaction without Authorize

paultheu:
When an OCPP 1.6 charger performs local authorization it skips the Authorize message and sends StartTransaction directly. The on_start_transaction handler now also sets charger-level (connector 0) id_tag and not only the connector-level id_tag

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction tracking for single-connector chargers by recording the authorization identifier for locally authorized charging sessions.
  * Prevented charger-level authorization data from being overwritten on multi-connector chargers.
  * Connector-specific authorization tracking continues to work as before.


drc38:
Why does the idtag need to be set at the charger level?

paultheu:
The only Home Assistant entity for idTag is the charger-level IdTag sensor. It is in CHARGER_ONLY, so it reads connector 0 — the slot written by Authorize.

OCPP 1.6 does not require Authorize before StartTransaction (LocalPreAuthorize / authorization cache). Without also writing connector 0 in StartTransaction, sensor.<cpid>_id_tag stays empty for local authorization even though the tag is already present on the connector.

I use that charger-level IdTag entity in automations (for example to set charger current).

drc38:
Please document details of this in the docs so that for multiple connectors its value is understood ie most recent connector idtag authorised on a transaction start

Alternatively, create an if statement to update station id tag only if connector count is 1

paultheu:
I opted for the if statement to prevent misunderstandings in case of multi-connector chargers

drc38 (approved):

Issue #1565, opened 2025-03-09
SetChargingProfile issue


npirtube:
**Describe the bug**
When I try to change the intensity of my charger it sends an error due to the type of profile sent.

**To Reproduce**
Try to change the maximum amperage value via the web interface.

**Expected behavior**
You should send an “Absolute” profile instead of “Relative”.
Also, it would be simpler to send a message by changing the key instead of a SetChargingProfile

**Screenshots**

![Image](https://github.com/user-attachments/assets/66063ee0-ffae-48e8-9572-7767a87ac35d)

`2025-03-09 20:11:37.829 INFO (MainThread) [ocpp] Plaza_1: send [2,"9706e144-0ceb-4349-8ed6-e83888e0d4c0","SetChargingProfile",{"connectorId":0,"csChargingProfiles":{"chargingProfileId":8,"stackLevel":1,"chargingProfileKind":"Relative","chargingProfilePurpose":"ChargePointMaxProfile","chargingSchedule":{"chargingRateUnit":"W","chargingSchedulePeriod":[{"startPeriod":0,"limit":22000}]}}}]
2025-03-09 20:11:38.087 INFO (MainThread) [ocpp] Plaza_1: receive message [4,"9706e144-0ceb-4349-8ed6-e83888e0d4c0","PropertyConstraintViolation","When chargingProfilePurpose =ChargingStationMaxProfile, chargingProfileKind SHALL NOT be Relative",{}]
2025-03-09 20:11:38.088 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [140528426986384] Unexpected exception
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 245, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2794, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2837, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1007, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2150, opened 2026-09-18
Add quirks implementation

jhenkens:
Autel has a ["broken" OCPP implementation](https://github.com/lbbrhzn/ocpp/issues/1565), but that is work-aroundable. More importantly, it actually "worked" just fine prior to the latest changes on `main` (https://github.com/lbbrhzn/ocpp/pull/2131), because it would fail the Max, then set a TxDefaultProfile. 

I am proposing that we introduce a quirks implementation which can augment the default OCPP behavior for a given manufacturer, overriding methods as needed.

The first implementation is specifically to correct this max-charging-behavior for Autel - changing the schedule from Relative to Absolute.

## Summary by CodeRabbit

- **New Features**
  - Added vendor-specific handling for Autel MaxiChargers.
  - Autel charging-rate requests now use absolute charging profiles with a standardized start time and requested limits.
  - Added support for the OCPP `startSchedule` option.
  - Vendor detection now applies the appropriate charging behavior automatically while preserving standard behavior for other vendors.


drc38:
Try out #2142 which adds a txprofile slider

jhenkens:
Yes, that can work too. But the integrated station-max slider would still remain broken for Autel chargers. In my opinion, broken but workarounds using different features is less than ideal, when it could be fixed. Downside, is more code. But by using a per-vendor, hopefully it doesn't make the code too confusing.

drc38:
Since the beginning of this repo it has been a conscious decision not to include workarounds for chargers that are clearly non-compliant with the specification. #2131 explains why the TxDefaultProfile is not a suitable fallback for the slider when there is more than 1 connector, if only 1 connector is present then the #2142 slider can be used.

jhenkens:
To each their own. I think that's a mistake that instead causes the same workaround to be implemented via documentation, and reduced features for users with certain chargers. This isn't some commercial CSMS where we might actually be able to convince Autel it's worth changing their firmware for us.

An alternative would be an advanced configuration to "use_absolute_range_for_charge_point_max_profile", and have the selection be user configured.

drc38:
> An alternative would be an advanced configuration to "use_absolute_range_for_charge_point_max_profile", and have the selection be user configured.

I'd be ok with that approach

PR #2154, opened 2026-09-19
Absolute charge point max

jhenkens:
See https://github.com/lbbrhzn/ocpp/pull/2150

Autel has a ["broken" OCPP implementation](https://github.com/lbbrhzn/ocpp/issues/1565), but that is work-aroundable. More importantly, it actually "worked" just fine prior to the latest changes on `main` (https://github.com/lbbrhzn/ocpp/pull/2131), because it would fail the Max, then set a TxDefaultProfile. 

Rather than implemented a manufacturer specific solution, we can implement a configuration option to change how the ChargePointMax is set - absolute or relative. 

This change is two commits, incase 2.0.1 changes are not desired. This change will be verified with an Autel charger on OCPP 1.6 by end of day.

## Summary by CodeRabbit

- **New Features**
  - Added an option to send station-wide charging limits as absolute schedules.
  - The setting is available during setup and in the options flow; relative schedules remain the default.
  - Absolute schedules use a fixed historical start time, while per-connector limits remain unchanged.

- **Bug Fixes**
  - Added configuration migration so existing installations receive the new setting automatically.

- **Documentation**
  - Documented compatibility guidance for Autel MaxiCharger devices and station-wide charging limits.


drc38 (approved):

jhenkens:
Thank you. Confirmed it works on 1.6 on my Autel charger.

